PR #18, opened 2024-07-03
[FCE-207] Fix error on muted track renegotiation

kamil-stasiak:
## Description

This PR fixes an error that occurs when a muted track is renegotiated.

## Motivation and Context

Steps to reproduce:
- Join the room with Peer 1.
- Join the room with Peer 2.
- Peer 2 mutes the microphone.
- Peer 2 turns off the camera.
- The error appears for Peer 1.

## Types of changes

- type `AudioOrVideoType` from `react-client` has been replaced by `TrackKind` from `ts-client`


packages/react-client/src/DeviceManager.ts
```diff
@@ -164,7 +164,7 @@ const handleNotAllowedError = async (constraints: MediaStreamConstraints): Promi
   return await getMedia({ video: false, audio: false }, { video: PERMISSION_DENIED, audio: PERMISSION_DENIED });
 };
 
-const getError = (result: GetMedia, type: AudioOrVideoType): DeviceError | null => {
+const getError = (result: GetMedia, type: TrackKind): DeviceError | null => {
```
karkakol, line 167: Imo `TrackType` sounds better than `TrackKind`

packages/ts-client/src/webrtc/transceivers.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      return acc
+    }, {} as Record<string, string>);
+
+
+export const getMuted = <EndpointMetadata, TrackMetadata>(
+  midToTrackId: Map<string, string> = new Map(),
+  localEndpoint: EndpointWithTrackContext<EndpointMetadata, TrackMetadata>,
+): Record<string, string> => {
+  return [...midToTrackId.entries()]
```
karkakol, line 42: midToTrackId.entries() would not be enough?

packages/ts-client/src/webrtc/transceivers.ts
```diff
@@ -0,0 +1,49 @@
...
+): Record<string, string> | null =>
+  connection.getTransceivers()
+    .filter((transceiver) => transceiver.sender.track?.id && transceiver.mid)
+    .reduce((acc, transceiver) => {
+      const localTrackId = transceiver.sender.track!.id;
+      const mid = transceiver!.mid!;
+
+      const trackContext = Array.from(localTrackIdToTrack.values()).find(
+        (trackContext) => trackContext?.track?.id === localTrackId,
+      )!;
```
Rados13, line 30: Could we move it to a function?

packages/ts-client/src/webrtc/transceivers.ts
```diff
@@ -0,0 +1,49 @@
...
+  localEndpoint: EndpointWithTrackContext<EndpointMetadata, TrackMetadata>,
+): Record<string, string> => {
+  return [...midToTrackId.entries()]
+    .filter(([_mid, trackId]) => localEndpoint.tracks.get(trackId))
+    .reduce((acc, [mid, trackId]) => {
+      acc[mid] = trackId;
+
+      return acc
+    }, {} as Record<string, string>);
+}
```
Rados13, line 49: If I understand correctly this function should return muted tracks, where do we filter out tracks that are not muted here?

packages/react-client/src/DeviceManager.ts
czerwiukk: maybe update the variable name as well to `kind: TrackKind`

karkakol: Agreed with @czerwiukk. lets rename function parameter name to `kind`.

packages/ts-client/src/webrtc/webRTCEndpoint.ts
```diff
@@ -1607,13 +1588,16 @@ export class WebRTCEndpoint<
...
+
       if (this.checkIfTrackBelongToEndpoint(trackId, this.localEndpoint))
         return;
+      if (!isTrackKind(event.track.kind)) throw new Error("Track has no kind")
 
       const trackContext = this.trackIdToTrack.get(trackId)!;
 
       trackContext.stream = stream;
       trackContext.track = event.track;
+      trackContext.trackKind = event.track.kind;
```
czerwiukk, line 1600: i may have missed something, but why are we copying the `event.track.kind` information to another field while we can access it through `trackCotnext.track.kind`?

packages/ts-client/src/webrtc/transceivers.ts
```diff
@@ -0,0 +1,49 @@
...
+): Record<string, string> | null => {
+  if (!connection) return null;
+
+  const active = getActive(connection, localTrackIdToTrack)
+  const muted = getMuted(midToTrackId, localEndpoint)
+
+  return { ...active, ...muted };
+};
+
+export const getActive = <EndpointMetadata, TrackMetadata>(
```
czerwiukk, line 18: what is active? isn't the name lacking something?

czerwiukk: looking at the code and comments, i'd go for getActiveTracks/mutedTracks

packages/ts-client/src/webrtc/transceivers.ts
kamil-stasiak: I need it to map the result of `entries()`, which is an `IterableIterator`, to a simple array.

packages/ts-client/src/webrtc/webRTCEndpoint.ts
kamil-stasiak: The `track` field is optional; it's possible to mute a track (replace the track with null) and then unmute it in the future. We need the `trackKind` to assess bandwidth even though the track is muted

packages/ts-client/src/webrtc/transceivers.ts
kamil-stasiak: I've changed the names of these functions in this file, and I've added some comments

packages/ts-client/src/webrtc/transceivers.ts
```diff
@@ -1,49 +1,59 @@
...
+): MidToTrackId | null => {
   if (!connection) return null;
 
-  const active = getActive(connection, localTrackIdToTrack)
-  const muted = getMuted(midToTrackId, localEndpoint)
+  // - negotiated unmuted tracks: tracks added in previous negotiation, data is being transmitted
+  // - not yet negotiated tracks: tracks added in this negotiation, data will be transmitted after successful negotiation
+  const transceiver = getTransceiverMapping(connection, localTrackIdToTrack)
+
+  // - negotiated unmuted tracks: tracks added in previous negotiation, data is being transmitted
```
Rados13, line 20: Why do we have the same comment two times?

kamil-stasiak: Because it's an intersection of these two functions

Rados13 (approved):

PR #33, opened 2024-07-09
FCE-263 Stop screen share on browser event

kamil-stasiak:
## Description

The screen share track is being removed on any stopped event for any `TrackType` (`audio`, `video`, or `audiovideo`).

## Motivation and Context

When the user stops screen sharing from the browser popup, the track is not removed.

## Types of changes

### DIFF
--- a/packages/react-client/src/useSetupMedia.tsx
+++ b/packages/react-client/src/useSetupMedia.tsx
@@ -312,9 +312,9 @@ export const createUseSetupMediaHook = <PeerMetadata, TrackMetadata>(
     useEffect(() => {
       const onScreenShareStop: ClientEvents<PeerMetadata, TrackMetadata>["deviceStopped"] = async (event, client) => {
         const stream = client.devices.screenShare.broadcast?.stream;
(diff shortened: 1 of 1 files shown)


packages/react-client/src/useSetupMedia.tsx
```diff
@@ -305,7 +312,7 @@ export const createUseSetupMediaHook = <PeerMetadata, TrackMetadata>(
     useEffect(() => {
       const onScreenShareStop: ClientEvents<PeerMetadata, TrackMetadata>["deviceStopped"] = async (event, client) => {
         const stream = client.devices.screenShare.broadcast?.stream;
```
czerwiukk, line 314: wouldn't it make more sense like that?
```
const isBroadcastedTrackStopped = client.status === "joined" && event.mediaDeviceType === "displayMedia" && !!stream;
if(!isBroadcastedTrackStopped) return;

await client.devices.screenShare.removeTrack();
```

czerwiukk (approved):

PR #41, opened 2024-07-18
Publish RTCStatsReports to Fishjam

Karolk99:
## Description

We want to gather rtc stats on the client side and send them to the Fishjam server for debugging.

## Types of changes

      not work as expected)


Karolk99:
Two questions remain:
* Should `RTCStatsReport` be sent using `MediaEvents`, or should it use a new notification like `RTCStatsReport`?
* Should inactive outbound tracks be sent?

packages/ts-client/src/FishjamClient.ts
```diff
@@ -714,6 +726,29 @@ export class FishjamClient<
...
+
+      const mediaEvent = generateMediaEvent('RTCStatsReport', tracksStatistics);
+      const serializedMediaEvent = serializeMediaEvent(mediaEvent);
+
+      const message = PeerMessage.encode({
+        mediaEvent: { data: serializedMediaEvent },
+      }).finish();
+
+      this.websocket?.send(message);
+    });
```
Rados13, line 749: Wouldn't it be simpler to just send all types of reports and then filter it on the backend side? Because currently if some user's browser has changed type we won't see it. Also if we would want to gather more statistics e.g: candidate statistics it will require modification in web-client-sdk.

packages/ts-client/src/FishjamClient.ts
```diff
@@ -714,6 +725,26 @@ export class FishjamClient<
...
+          tracksStatistics[key] = report;
+      });
+
+      const message = PeerMessage.encode({
+        rtcStatsReport: { data: JSON.stringify(tracksStatistics) },
+      }).finish();
+
+      this.websocket?.send(message);
+    });
+  }
```
kamil-stasiak, line 746: You don't need `null` in `getStatistics`. That function could be async to remove the unnecessary `.then()` callback. You could also avoid mutating `trackStatistics` by using a reduce function.

```suggestion
private async sendStatistics() {
    const statistics = await this.getStatistics()

    const tracksStatistics = Object.entries(statistics)
      .filter(([_key, report]) => report.type === 'inbound-rtp' || report.type === 'outbound-rtp')
      .reduce((acc, [key, report]) => ({ ...acc, [key]: report }),
        {} as Record<string, RTCInboundRtpStreamStats | RTCOutboundRtpStreamStats>);

    const message = PeerMessage.encode({
      rtcStatsReport: { data: JSON.stringify(tracksStatistics) },
    }).finish();

    this.websocket?.send(message);
  }
```

kamil-stasiak (changes requested):

kamil-stasiak (approved):

PR #54, opened 2024-08-12
Set printWidth to 120

kamil-stasiak:
## Description

Set Prettier's `printWidth` to 120 for `ts-client`


packages/ts-client/.prettierrc
```diff
@@ -6,5 +6,6 @@
   "singleQuote": true,
   "tabWidth": 2,
   "trailingComma": "all",
-  "useTabs": false
+  "useTabs": false,
+  "printWidth": 120
```
mironiasty, line 10: super nit: it should go under `bracketSameLine` (property sorting)

mironiasty (approved):

PR #55, opened 2024-08-12
FCE-361 Remove metadata from updateTrack

kamil-stasiak:
## Description

Remove `newMetadata` from `replaceTrack` along with the code responsible for handling it.

## Motivation and Context

The `replaceTrack` function is currently not atomic because it first replaces the track and then updates the metadata, which is unnecessary. We could remove this parameter to simplify the code. Without this parameter, the user can chain `replaceTrack` with `updateTrackMetadata`.

## Types of changes


czerwiukk (approved):
looks good, i suppose this is a tidying-up groundwork for removing the metadata types in ts and react client?

packages/ts-client/src/webrtc/tracks/LocalTrackManager.ts
```diff
@@ -100,20 +100,18 @@ export class LocalTrackManager<EndpointMetadata, TrackMetadata> {
...
     newTrack: MediaStreamTrack | null,
-    // todo remove webrtc with newTrackMetadata
-    newTrackMetadata?: TrackMetadata,
   ): Promise<void> => {
     this.ongoingTrackReplacement = true;
     try {
-      await this.local.replaceTrack(webrtc, trackId, newTrack, newTrackMetadata);
+      await this.local.replaceTrack(webrtc, trackId, newTrack);
     } catch (_error) {
+      /* empty */
```
mironiasty, line 111: Why does `local.replaceTrack` throw exception, just to be ignored by `replaceTrackHandler`?

mironiasty (approved):

PR #56, opened 2024-08-12
FCE-362 signaling as a string

kamil-stasiak:
## Description

Refactor signaling to a single string.

## Motivation and Context

The signaling URL as an object was very confusing.

## Types of changes


examples/react-client/minimal-react/src/components/App.tsx
```diff
@@ -52,6 +52,7 @@ export const App = () => {
             connect({
               peerMetadata: { name: "John Doe" }, // example metadata
               token: token,
+              url: "ws://localhost:5002",
```
czerwiukk, line 55: **suggestion (non-blocking):** consider moving the string to a variable

packages/ts-client/src/FishjamClient.ts
```diff
@@ -354,16 +337,17 @@ export class FishjamClient<PeerMetadata, TrackMetadata> extends (EventEmitter as
     this.status = 'initialized';
   }
 
+  private getUrl(url: string) {
+    if (url.endsWith('/')) return url + WEBSOCKET_PATH;
+    return url + '/' + WEBSOCKET_PATH;
```
czerwiukk, line 342: **suggestion:** consider sticking to template literals while working with strings in js as it is a safer operation
```suggestion
    if (url.endsWith('/')) return `${url}${WEBSOCKET_PATH}`
    return `${url}/${WEBSOCKET_PATH}`
```

czerwiukk (approved):

PR #64, opened 2024-08-20
update screenshare types

czerwiukk:
## Description

Add missing types and remove stale ones.


packages/react-client/src/types.ts
```diff
@@ -181,7 +181,11 @@ export type UserMediaAPI<TrackMetadata> = {
 };
 
 export type ScreenshareApi<TrackMetadata> = {
-  startStreaming: (props?: { metadata?: TrackMetadata; requestAudio?: boolean }) => Promise<void>;
+  startStreaming: (props?: {
```
kamil-stasiak, line 184: I would use this in `screenShareTrackManager.ts`

```
const startStreaming: ScreenshareApi<TrackMetadata>["startStreaming"] = async (props) => {
```

kamil-stasiak (approved):

PR #67, opened 2024-08-23
Update useParticipiants to show all video tracks

mironiasty:
## Description

For random reason, `useParticipiants` only returned single audio/video track per peer. This change updates code so it returns all tracks. 

I also refactored videoroom example, to extract `App` into smaller components.

## Motivation and Context

Fix bug, that prevented testing videoshare

## Types of changes

      not work as expected)


examples/react-client/videoroom/src/components/VideoTracks.tsx
```diff
@@ -0,0 +1,30 @@
+import VideoPlayer from "./VideoPlayer";
+import { Track } from "@fishjam-cloud/react-client";
+
+type Props = {
+  id: string;
+  name: string;
+  videoTracks: Track<unknown>[] | undefined;
```
czerwiukk, line 7: **question:** what's the reason behind not using the optional parameter syntax?

czerwiukk: Anyways - is there a case in which `videoTracks` are going to be `undefined`? I think the SDK should return at least an empty list.

mironiasty: There is important difference between `videoTracks: Track<unknown>[] | undefined` and `videoTracks?: Track<unknown>[]`. So they can't be used interchangable.
Anyway, I removed `undefined` as it is indeed not required.

czerwiukk (approved):

kamil-stasiak (approved):

PR #68, opened 2024-08-23
Fce 394/track middleware

czerwiukk:
## Description

Adds track middleware to screenshare and video/audio managers.
Additionally, the track manager has been rewritten from a class to react hook.

## Motivation and Context

Track middleware allows easy manipulation of the tracks managed by the SDK before streaming them out.

I also rewritten the track manager to a hook, because I needed a state to store the middleware.
Adding a property to the TrackManager class wasn't enough, I also needed to synchronize with the react state.
Instead of pursuing the event emitting pattern, I decided to introduce an internal hook.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/Client.ts
+++ b/packages/react-client/src/Client.ts
@@ -42,7 +42,7 @@ export type ClientApi<PeerMetadata, TrackMetadata> = {
   bandwidthEstimation: bigint;
   status: PeerStatus;
   media: MediaState | null;
-  devices: Devices<TrackMetadata>;
+  devices: Devices;
 
   isReconnecting: () => boolean;
 };
@@ -313,7 +313,7 @@ export class Client<PeerMetadata, TrackMetadata> extends (EventEmitter as {
   public bandwidthEstimation: bigint = BigInt(0);
   public status: PeerStatus = null;
   public media: MediaState | null = null;
-  public devices: Devices<TrackMetadata>;
+  public devices: Devices;
 
   public reconnectionStatus: ReconnectionStatus = "idle";
 
@@ -829,7 +829,7 @@ export class Client<PeerMetadata, TrackMetadata> extends (EventEmitter as {
       localTracks[track.trackId] = this.trackContextToTrack(track);
     });
 
-    const devices: Devices<TrackMetadata> = {
+    const devices: Devices = {
       camera: {
         status: deviceManagerSnapshot?.video?.devicesStatus || null,
         stream: deviceManagerSnapshot?.video.media?.stream || null,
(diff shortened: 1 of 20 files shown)


packages/react-client/src/Client.ts
```diff
@@ -397,17 +385,6 @@ export class Client<PeerMetadata, TrackMetadata> extends (EventEmitter as {
       this.emit("authError", reason, this);
     });
 
-    this.tsClient.on("disconnected", () => {
```
czerwiukk, line 400: **thought:** i think it'd be good to just call state to snapshot instead of removing the whole handler

packages/react-client/src/screenShareTrackManager.ts
```diff
@@ -29,12 +29,37 @@ export const useScreenShare = <PeerMetadata, TrackMetadata>(
...
+    if (!state?.stream) return;
+
+    const [videoTrack, audioTrack] = getTracks(state.stream);
+
+    if (!middleware) {
+      return await replaceTracks(videoTrack, audioTrack);
+    }
+
+    const [newVideoTrack, newAudioTrack] = middleware(videoTrack, audioTrack);
+    await replaceTracks(newVideoTrack, newAudioTrack);
```
kamil-stasiak, line 54: Just an idea:
```
    const [newVideoTrack, newAudioTrack] =  middleware ? middleware(videoTrack, audioTrack) : [videoTrack, audioTrack]
    await replaceTracks(newVideoTrack, newAudioTrack);
```

packages/react-client/src/trackManager.ts
```diff
@@ -1,111 +1,141 @@
...
-  };
+    setCurrentTrackMiddleware(() => middleware);
+  }
 
-  public startStreaming = async (
+  function getCurrentTrack(): Track<TrackMetadata> | null {
+    return getRemoteOrLocalTrack<PeerMetadata, TrackMetadata>(tsClient, currentTrackId);
+  }
+
+  async function initialize(deviceId?: string) {
```
kamil-stasiak, line 57: nit: In the future, we need to decide on a consistent code style for functions (fat arrow or function keyword). Can we refactor this file to use only one style?

kamil-stasiak (approved):

packages/react-client/src/screenShareTrackManager.ts
czerwiukk: i like that and i will take it even further

packages/react-client/src/trackManager.ts
czerwiukk: converted to `function`

kamil-stasiak (approved):

kamil-stasiak (approved):

PR #71, opened 2024-08-26
[FCE-399] Remove create fn

czerwiukk:
## Description

Removes the create function and segregates the hooks in the `/hooks` directory.

## Motivation and Context

We don't want to export our hooks through a function. The boundary of a client is now managed by the FishjamProvider.

## Types of changes

      not work as expected)


packages/react-client/src/fishjamProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function FishjamProvider({ children, config, deviceManagerDefaultConfig }: FishjamProviderProps) {
+  const client = useMemo(
+    () =>
+      new Client({
+        clientConfig: config,
+        deviceManagerDefaultConfig,
+      }),
+    [config, deviceManagerDefaultConfig],
```
czerwiukk, line 22: **thought:** potentially dangerous: changing config may erase the state

packages/react-client/src/hooks/setupMedia.ts
```diff
@@ -0,0 +1,254 @@
...
+      storage: config.storage,
+    });
+  }, [config, state.client]);
+
+  useEffect(() => {
+    if (!configRef.current.startOnMount) return;
+    if (
+      state.client.audioDeviceManager.getStatus() === "uninitialized" ||
+      state.client.videoDeviceManager.getStatus() === "uninitialized"
+    ) {
```
kamil-stasiak, line 30: I've removed this if and I've added guard to `initializeDevices` method

kamil-stasiak (approved):

PR #73, opened 2024-08-26
Update package configs for better TS

mironiasty:
## Description

- Update project configuration to have TypeScript checks configured in similar way as in Mobile repo.
- Make sure all packages are checked when running global command (few were missing  :( )
- Update subproject names to match convention
- Small type update, to fix TS error in Videoroom example

## Motivation and Context

Fix bugs and prevent them in future

## Types of changes

      not work as expected)


packages/react-client/src/types.ts
```diff
@@ -255,10 +256,7 @@ export type CreateFishjamClient<PeerMetadata, TrackMetadata> = {
   useMicrophone: () => Devices["microphone"] & TrackManager<TrackMetadata>;
   useClient: () => Client<PeerMetadata, TrackMetadata>;
   useReconnection: () => UseReconnection;
-  useParticipants: () => {
-    localParticipant: PeerStateWithTracks<PeerMetadata, TrackMetadata> | null;
-    participants: PeerStateWithTracks<PeerMetadata, TrackMetadata>[];
-  };
+  useParticipants: () => Participiants<PeerMetadata, TrackMetadata>;
```
mironiasty, line 259: TS complain that return type of this function was not exported.

PR #77, opened 2024-08-30
Remove Client.ts

czerwiukk:
## Description

This pull request removes the mediator Client class and moves the functionality into smaller React Hooks.
Additionally, we decided to remove the `useSetupMedia` hook.

## Motivation and Context

The main motivation is to remove the Client class from react client. We only need to synchronize the state with the TS Client. All additional state can be already maintained in the react hooks.

## Types of changes

      not work as expected)


packages/react-client/src/hooks/public.ts
```diff
@@ -1,6 +1,9 @@
...
-export { useCamera, useMicrophone } from "./devices";
+export { useConnect, useDisconnect } from "./connection";
+export { useReconnection } from "./reconnection";
+export { useCamera, useMicrophone, useInitializeDevices } from "./devices";
 export { useParticipants } from "./participants";
 export { useScreenShare } from "./screenShare";
-export { useSelector, useStatus, useTracks, useClient } from "./state";
-export { useSetupMedia } from "./setupMedia";
+export { useStatus } from "./state";
+export { useFishjamClient } from "./tsClient";
```
mironiasty, line 7: Do we need to export this hook?

packages/react-client/src/hooks/tsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+  return {
+    rawMetadata: peer.rawMetadata,
+    metadata: peer.metadata,
+    metadataParsingError: peer.metadataParsingError,
+    id: peer.id,
+    tracks,
+  };
+}
+
+export const useFishjamClient = () => {
```
mironiasty, line 97: Let's remove this export from API in future ticket, and for now it could be marked as deprecated
```suggestion
/**
 * @deprecated WIll be removed from public API in  FCE-XXX
 */
export const useFishjamClient = () => {
```

packages/react-client/src/hooks/devices.ts
```diff
@@ -1,33 +1,98 @@
...
+  const devices = deviceState.devices ?? [];
+  const activeDevice = deviceState.media?.deviceInfo ?? null;
 
   return { ...trackManager, streamedTrack, streamedTrackId, isAudioPlaying, stream, devices, activeDevice };
 }
+
+type InitializeDevicesProps = { autoInitialize?: boolean };
+
+export const useInitializeDevices = (props?: InitializeDevicesProps) => {
+  const areDevicesInitialized = useRef(false);
```
kamil-stasiak, line 44: This hook works similar to just deleted `useSetupMedia`. The user can use it in many different components, and by doing so, the device manager will start multiple times, which I don't think we are able to handle. I think `areDevicesInitialized` should be part of a shared state, or `useInitializeDevices` shouldn't be a hook at all, just a simple function (in that case, areDevicesInitialized also needs to be in shared state).

packages/react-client/src/hooks/screenShare.ts
```diff
@@ -14,7 +14,8 @@ export const useScreenShare = (): ScreenshareApi => {
   const ctx = useFishjamContext();
 
   const [state, setState] = ctx.screenshareState;
-  const tsClient = ctx.state.client.getTsClient();
+  const { fishjamClientRef } = useFishjamContext();
+  const tsClient = useMemo(() => fishjamClientRef.current, [fishjamClientRef]);
```
mironiasty, line 18: Is this memo required here? 🤔

packages/react-client/src/hooks/reconnection.ts
```diff
@@ -0,0 +1,41 @@
...
+      client.off("reconnected", setIdle);
+      client.off("reconnectionRetriesLimitReached", setError);
+    };
+  }, [fishjamClientRef]);
+
+  return {
+    status: reconnectionStatus,
+    isReconnecting: reconnectionStatus === "reconnecting",
+    isError: reconnectionStatus === "error",
+    isIdle: reconnectionStatus === "idle",
```
mironiasty, line 39: Do we need these variables?

packages/react-client/src/hooks/devices.ts
```diff
@@ -1,33 +1,98 @@
...
+  }, [videoDeviceManagerRef, audioDeviceManagerRef]);
+
+  const autoInitialize = Boolean(props?.autoInitialize);
+
+  useEffect(() => {
+    if (!autoInitialize) return;
+    initializeDevices();
+  }, [initializeDevices, autoInitialize]);
+
+  return { initializeDevices };
```
kamil-stasiak, line 97: **Future improvement idea**: WDYT about adding additional state like `initializing` that will inform the user that the manager has started, so the user can disable buttons or display some kind of popup like "Grant access to camera and microphone"?

Of course in separate PR

packages/react-client/src/hooks/devices.ts
```diff
@@ -1,33 +1,98 @@
...
   const stream = currentTrack?.stream ?? null;
-  const devices = state.devices.microphone.devices ?? [];
-  const activeDevice = state.devices.microphone.deviceInfo;
+  const devices = deviceState.devices ?? [];
+  const activeDevice = deviceState.media?.deviceInfo ?? null;
 
   return { ...trackManager, streamedTrack, streamedTrackId, isAudioPlaying, stream, devices, activeDevice };
 }
+
+type InitializeDevicesProps = { autoInitialize?: boolean };
```
mironiasty, line 41: Maybe not for this ticket, but... do we need this flag? 🤔

packages/react-client/src/hooks/devices.ts
```diff
@@ -1,33 +1,98 @@
...
+
+export const useInitializeDevices = (props?: InitializeDevicesProps) => {
```
kamil-stasiak, line 43: **Future improvement idea:** I think it would be beneficial for the user to be allowed to request access to only the camera or the microphone. By default, it should start both, as you have implemented.

packages/react-client/src/hooks/deviceManagers/audioDeviceManager.ts
mironiasty, line 1: maybe this (and below) file should be called `useAudioDeviceManager.ts`?

packages/react-client/src/hooks/deviceManagers/audioDeviceManager.ts
```diff
@@ -0,0 +1,9 @@
+import { useFishjamContext } from "../fishjamContext";
+import { useDeviceManager } from "./deviceManager";
+
+export const useAudioDeviceManager = () => {
+  const { audioDeviceManagerRef } = useFishjamContext();
+  const audioDeviceState = useDeviceManager(audioDeviceManagerRef.current);
+
+  return { ...audioDeviceState };
```
mironiasty, line 8: Why not just:
```suggestion
  return audioDeviceState;
```

mironiasty (commented):
I would prefer to follow `useHookName.ts` naming strategy. It seems to be more popular (example: https://github.com/software-mansion/react-native-reanimated/tree/main/packages/react-native-reanimated/src/hook)

packages/react-client/src/hooks/peerStatus.ts
```diff
@@ -0,0 +1,45 @@
...
+    };
+    const setJoined = () => {
+      setPeerStatus("joined");
+    };
+    const setDisconnected = () => {
+      setPeerStatus(null);
+    };
+    const setConnected = () => {
+      setPeerStatus("connected");
+    };
```
kamil-stasiak, line 24: **Discussion idea**: Based on my experience, `joined` and `connected` are very confusing. I can't count how many times I made a mistake and considered the state `connected` as final and correct. The correct state is `joined`. I suggest in the future to replace `joined` with `connected`, but I have no idea what to call the old `connected`

packages/react-client/src/hooks/reconnection.ts
czerwiukk: agreed to simplify

packages/react-client/src/hooks/tsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+    disconnect,
+    removeTrack,
+    replaceTrack,
+    getStatistics,
+    getBandwidthEstimation,
+    setTrackBandwidth,
+    setEncodingBandwidth,
+    setTargetTrackEncoding,
+    enableTrackEncoding,
+    disableTrackEncoding,
```
kamil-stasiak, line 209: **Discussion Idea**: If we decide to always use `useCamera`, `useMicrophone`, and `useScreenShare`, shouldn't we remove those methods?

packages/react-client/src/hooks/devices.ts
czerwiukk: I thought we might provide two options to handle this:
1. Doing this automatically - just invoke the hook in the component and the browser will simply ask the user once the component rendered.
2. Doing this manually - the devs might want to invoke the method manually upon handling some action - they could do that by using the method returned by the hook.

I decided the latter one to be the default behavior, nothing happens automatically until verbosely stated.
What do you think?

packages/react-client/src/hooks/deviceManagers/audioDeviceManager.ts
czerwiukk: i think it's a leftover from extending the object by some properties

packages/react-client/src/hooks/connection.ts
```diff
@@ -1,36 +1,24 @@
...
-  const { state }: FishjamContextType = useFishjamContext();
+  const client = useFishjamClient();
 
   return useMemo(() => {
     return (config) => {
-      state.client.connect(config);
+      client.connect(config);
       return () => {
-        state.client.disconnect();
+        client.disconnect();
```
kamil-stasiak, line 12: **Discussion Idea**: If we have `useDisconnect`, should we return anything from the `connect()` function? Maybe we could return a promise that resolves when the user establishes a connection.

kamil-stasiak (approved):

czerwiukk:
> I would prefer to follow `useHookName.ts` naming strategy. It seems to be more popular (example: https://github.com/software-mansion/react-native-reanimated/tree/main/packages/react-native-reanimated/src/hook)

I feel the `use` is a bit redundant, though I see the benefit of recognizing the hook file at the first glance.
I'm fine with any of those approaches. We can take it on in any current / upcoming projects. Maybe there's some eslint rule?

kamil-stasiak:
I've written a few comments, but only this one is important to address: https://github.com/fishjam-cloud/web-client-sdk/pull/77#discussion_r1741628365. The others are just ideas

packages/react-client/src/hooks/tsClient.ts
czerwiukk: The `useFishjamClient` hook is now exported as public, but as you can see above, we decided to hide it after sorting out how to handle tests. 
In this case we might leave those methods as the hook would be an internal one.

packages/react-client/src/hooks/peerStatus.ts
czerwiukk: Makes sense, we could have us another "proposal" ticket.

packages/react-client/src/hooks/devices.ts
czerwiukk: My bad, the ref should be kept in the provider in this case. thanks

packages/react-client/src/hooks/connection.ts
czerwiukk: big up, i see no benefit from returning the disconnect fn, then some benefit of making the connect awaitable

packages/react-client/src/hooks/devices.ts
kamil-stasiak: I like @czerwiukk's approach.

packages/react-client/src/hooks/devices.ts
mironiasty: Sound like a plan! :)

packages/react-client/src/hooks/screenShare.ts
czerwiukk: reasumując wszystkie aspekty kwintesencji tematu, dochodzę do fundamentalnej konkluzji - nie jest

packages/react-client/src/hooks/devices.ts
kamil-stasiak: https://linear.app/swmansion/issue/FCE-456/return-device-manager-status-from-useinitializedevices

packages/react-client/src/hooks/devices.ts
kamil-stasiak: https://linear.app/swmansion/issue/FCE-457/allow-user-to-only-initialise-audio-or-video

packages/react-client/src/hooks/peerStatus.ts
kamil-stasiak: https://linear.app/swmansion/issue/FCE-458/rename-client-statuses-joined-connected

packages/react-client/src/hooks/tsClient.ts
kamil-stasiak: https://linear.app/swmansion/issue/FCE-458/rename-client-statuses-joined-connected

packages/react-client/src/hooks/connection.ts
kamil-stasiak: https://linear.app/swmansion/issue/FCE-460/return-promise-from-connect-function

PR #82, opened 2024-09-03
FCE-441 Screen doesn't work

kamil-stasiak:
## Description

- Media manager properly saves last selected device to local storage. 
- Track metadata was added to `useScreenShare`.
- Temporary solution: `useCamera` and `useMicrophone` returns local stream 

## Motivation and Context

Screen share in videoroom does not work.

## Types of changes


packages/react-client/src/DeviceManager.ts
```diff
@@ -47,16 +47,16 @@ export class DeviceManager
...
     this.deviceType = deviceType;
 
     this.constraints = toMediaTrackConstraints(defaultConfig?.trackConstraints ?? true);
   }
 
-  private createStorageConfig(storage?: boolean | StorageConfig): StorageConfig | null {
+  private createStorageConfig(deviceType: "audio" | "video", storage?: boolean | StorageConfig): StorageConfig | null {
     if (storage === false) return null;
-    if (storage === true || storage === undefined) return getLocalStorageConfig(this.deviceType);
+    if (storage === true || storage === undefined) return getLocalStorageConfig(deviceType);
```
czerwiukk, line 59: **question:** why is this change needed? if the device type was undefined, then maybe we should set it before invoking `this.createStorageConfig` in line 50.

packages/react-client/src/hooks/screenShare.ts
```diff
@@ -23,8 +25,26 @@ export const useScreenShare = (): ScreenshareApi => {
     });
     const [video, audio] = getTracks(stream);
 
-    const addTrackPromises = [tsClient.addTrack(video)];
-    if (audio) addTrackPromises.push(tsClient.addTrack(audio));
+    const addTrackPromises = [
+      tsClient.addTrack(video).then(async (trackId) => {
```
czerwiukk, line 29: **question:** can't we just go with
```
tsClient.addTrack(video, { type: "screenShareVideo", paused: false })
```
?

packages/react-client/src/DeviceManager.ts
kamil-stasiak: Yeah, you're right. I thought about it, and I think this approach is safer because it's not dependent on the order of function invocations in the constructor. It makes createStorageConfig a pure function, so I decided to write it this way. TS disappointed me because it allows private readonly deviceType: "audio" | "video" to be undefined.

packages/react-client/src/hooks/screenShare.ts
kamil-stasiak: Yeah, you're right. I forgot that this is possible 😅

packages/react-client/src/DeviceManager.ts
czerwiukk: spox

czerwiukk (approved):

PR #85, opened 2024-09-04
FCE-449 Stream API proposal

czerwiukk:
## Description

The PR simplifies the Device API.
Now, we expose just one `stream` property.
It contains the middleware-processed stream straight from the device.
Then, while connected, the stream changes to the one that fishjam client is streaming to the web.

## Motivation and Context

Besides making the API simpler, I've also stumbled upon the need to introduce a "cleanup" function to the middleware feature. Spawning more than one blur processor resulted severe performance drops. Spawning one processor for device stream, then another one for the "streamed" stream wasn't possible without a cleanup option.

## Types of changes

      not work as expected)


packages/react-client/src/types.ts
```diff
@@ -169,9 +169,8 @@ export type ScreenshareState = {
 } | null;
 
 export type Device = {
-  streamedTrack: MediaStreamTrack | null;
-  streamedTrackId: TrackId | null;
-  stream: MediaStream | null;
+  currentlyStreamed: { track: MediaStreamTrack; trackId: TrackId; stream: MediaStream } | null;
+  rawStream: MediaStream | null;
```
kamil-stasiak, line 173: I'm not sure if `currentlyStreamed` is the right name. It's descriptive but a little bit odd. Since it's our public API, the name will stick with us. My suggestion is to use `participantStream` or `broadcastStream` instead. @mironiasty WDYT?

czerwiukk: `broadcastStream.stream` and `broadcastStream.track`, or `broadcast.stream` and `broadcast.track`?

mironiasty: duh, looks like I didn't publish my previous comment before

kamil-stasiak: To remove `rawStream` and `currentlyStreamed` we could implement one stream (with middleware) and additional field as an indicator:
https://linear.app/swmansion/issue/FCE-450/expose-local-stream-with-middleware

czerwiukk: > To remove `rawStream` and `currentlyStreamed` we could implement one stream (with middleware)
I will take time to try to achieve that in the simplest way.

packages/react-client/src/hooks/useDevices.ts
```diff
@@ -1,41 +1,62 @@
...
-  } = useFishjamContext();
-
+  const { videoTrackManager } = useFishjamContext();
   const { deviceState } = useVideoDeviceManager();
+  const { currentTrack, ...trackManager } = videoTrackManager;
+
+  const processedPreviewStream = useProcessedPreviewStream(videoTrackManager, deviceState.media?.track);
 
-  return { ...trackManager, ...getDeviceProperties(currentTrack, deviceState) };
+  const stream = currentTrack?.stream ?? processedPreviewStream ?? deviceState.media?.stream ?? null;
```
kamil-stasiak, line 16: Do we need to check `currentTrack?.stream`? 

```suggestion
  const stream = processedPreviewStream ?? deviceState.media?.stream ?? null;
```

packages/react-client/src/hooks/useDevices.ts
```diff
@@ -1,41 +1,62 @@
...
-    audioTrackManager: { currentTrack, ...trackManager },
-  } = useFishjamContext();
+  const { audioTrackManager } = useFishjamContext();
 
   const { deviceState } = useAudioDeviceManager();
+  const { currentTrack, ...trackManager } = audioTrackManager;
+
+  const processedPreviewStream = useProcessedPreviewStream(audioTrackManager, deviceState.media?.track);
+
+  const stream = currentTrack?.stream ?? processedPreviewStream ?? deviceState.media?.stream ?? null;
```
kamil-stasiak, line 42: Do we need to check `currentTrack?.stream`?

```suggestion
  const stream = processedPreviewStream ?? deviceState.media?.stream ?? null;
```

packages/react-client/src/hooks/useScreenShare.ts
```diff
@@ -49,17 +59,29 @@ export const useScreenShare = (): ScreenshareApi => {
...
+    cleanMiddleware();
+
+    if (middleware) {
+      const { videoTrack, audioTrack, onClear } = middleware(currentVideo, currentAudio);
+      cleanMiddlewareFnRef.current = onClear;
+      await replaceTracks(videoTrack, audioTrack);
+    }
 
-    await replaceTracks(newVideoTrack, newAudioTrack);
+    await replaceTracks(currentVideo, currentAudio);
```
kamil-stasiak, line 80: nit: this implementation could look like one in `startStreaming`. 
```ts
let [video, audio] = getTracks(stream);

if (middleware) {
  const { videoTrack, audioTrack, onClear } = middleware(video, audio);
  video = videoTrack;
  audio = audioTrack;
  cleanMiddlewareFnRef.current = onClear;
}

await replaceTracks(video, audio);
```

Or simplify it even more:

```
const [currentVideo, currentAudio] = getTracks(state.stream);

cleanMiddleware();

const { videoTrack, audioTrack, onClear } = middleware 
  ? middleware(currentVideo, currentAudio)
  : { videoTrack: currentVideo, audioTrack: currentAudio, onClear: null }
 
cleanMiddlewareFnRef.current = onClear;
await replaceTracks(videoTrack, audioTrack);
```

packages/react-client/src/hooks/useScreenShare.ts
```diff
@@ -74,11 +96,12 @@ export const useScreenShare = (): ScreenshareApi => {
 
     await Promise.all(removeTrackPromises);
 
-    setState(null);
-  }, [state, fishjamClientRef, setState]);
+    cleanMiddleware();
```
kamil-stasiak, line 99: In this implementation, we don't have the ability to `stopStreaming` without stopping the screen capturer. This is possible for the microphone and camera. Is this intended behaviour? If so, we could create another ticket for this.

packages/react-client/src/hooks/useTrackManager.ts
```diff
@@ -33,25 +35,39 @@ export const useTrackManager = ({ mediaManager, tsClient }: TrackManagerConfig):
...
-    setCurrentTrackMiddleware(() => middleware);
+    await tsClient.replaceTrack(currentTrack.trackId, track);
   }
 
   async function initialize(deviceId?: string) {
     await mediaManager?.start(deviceId ?? true);
     if (!currentTrackId) return;
     const newTrack = mediaManager.getTracks()[0];
-    await tsClient.replaceTrack(currentTrackId, newTrack);
+    await tsClient.replaceTrack(currentTrackId, clearAndGetProcessedTrack(newTrack));
```
kamil-stasiak, line 66: It replaces remote track, but what about local track? Would on preview screen?

packages/react-client/src/hooks/useTrackManager.ts
```diff
@@ -86,13 +104,12 @@ export const useTrackManager = ({ mediaManager, tsClient }: TrackManagerConfig):
...
     if (!newTrack) throw Error("New track is empty");
-
-    return tsClient.replaceTrack(currentTrack.trackId, newTrack);
+    return tsClient.replaceTrack(currentTrack.trackId, clearAndGetProcessedTrack(newTrack));
   }
 
   async function stopStreaming() {
     if (!currentTrack) return;
-
+    clearMiddleware();
```
kamil-stasiak, line 112: I don't think, that we should clear the middleware inside `stopStreaming` function. I understand that this invocation is mandatory because if the user stops, the streaming middleware is handled in a different place.

kamil-stasiak (commented):
The middleware logic is implemented in `useTrackManager`, `useScreenShare`, and `useProcessedPreviewStream`. 

It's okay to implement it in `useScreenShare` because it's a special case. But I think we should somehow merge the other two implementations.

Currently, if someone adds a blurred video, the middleware restarts. It's not a big deal, but we could avoid it. Probably, if we move the middleware logic to `DeviceManager`, the user would always get the correct track (with or without the middleware) from `getMedia()`. We could potentially remove `useProcessedPreviewStream` and simplify `useTrackManager` because it should only proxy the `setTrackMiddleware` invocation to the `DeviceManager`.

packages/react-client/src/hooks/useTrackManager.ts
czerwiukk: the preview one is processed in a different place, but i agree to unify

packages/react-client/src/hooks/useTrackManager.ts
czerwiukk: The stream is dropped either way when we stop streaming, so we need to clean up.
The middleware can be called again with another stream, but each call needs to have a corresponding cleanup.

packages/react-client/src/hooks/useDevices.ts
czerwiukk: That's the stream provided by ts client. `processedPreviewStream` is the processed one from the device

czerwiukk: Do you think we could ditch using the "re-packed" stream from tsclient and just leave the one straight from the device?

packages/react-client/src/hooks/useScreenShare.ts
czerwiukk: So far this is an intended behavior. Stopping streaming without stopping the capturer will result in browser telling the end user he is in fact sharing his tab.

czerwiukk:
> Currently, if someone adds a blurred video, the middleware restarts. It's not a big deal, but we could avoid it. Probably, if we move the middleware logic to `DeviceManager`.

This is correct. We can go for that improvement, but I'd like to do that in a separate PR. Let's ship the value and iterate.

kamil-stasiak (approved):

PR #87, opened 2024-09-05
FCE-453 hide fishjam client hook

czerwiukk:
## Description

Change useFishjamClient name to `useINTERNAL_FishjamClient`.

## Motivation and Context

Make it more obscure so the users dont use it.
## Types of changes

      not work as expected)


packages/react-client/src/hooks/public.ts
```diff
@@ -3,11 +3,7 @@ export { useReconnection } from "./useReconnection";
 export { useCamera, useMicrophone, useInitializeDevices } from "./useDevices";
 export { useParticipants } from "./useParticipants";
 export { useScreenShare } from "./useScreenShare";
-export { useStatus } from "./useState";
+export { useStatus } from "./useStatus";
+export { useINTERNAL_FishjamClient } from "./useFishjamClient";
```
kamil-stasiak, line 7: Idea: It's possible to export this hook from:
```ts
import { useINTERNAL_FishjamClient } from '@fishjam-cloud/react-client/internal';  
```
instead of

```ts
import { useINTERNAL_FishjamClient } from '@fishjam-cloud/react-client';  
```

WDYT? Is it worth investing additional time to make this change?

packages/react-client/src/hooks/useFishjamClient.ts
```diff
@@ -94,7 +94,11 @@ function endpointToPeerState(
   };
 }
 
-export const useFishjamClient = () => {
+/*
+This is an internally used hook.
+It is not meant to be used by the end user.
+*/
+export const useINTERNAL_FishjamClient = () => {
```
mironiasty, line 101: Maaaaybe:
```suggestion
export const useFishjamClient_DONOTUSE = () => {
```

packages/react-client/src/hooks/public.ts
mironiasty: If you know how to make it work that way - then go for it.

packages/react-client/src/hooks/useFishjamClient.ts
czerwiukk: only thing i dont like about it is that the words blend in uppercase

packages/react-client/src/hooks/public.ts
czerwiukk: done

packages/react-client/src/hooks/useFishjamClient.ts
kamil-stasiak: I don't have any preferences, but if we use `_` in the middle, we could also write is as `useFishjamClient_DO_NOT_USE`

mironiasty: `useFishjamClient_DO_NOT_USE` is also fine, let's change it and then it is ready to merge :)

mironiasty (approved):

PR #90, opened 2024-09-05
FCE-484 initializeDevices starts two streams

kamil-stasiak:
## Description

`initializeDevices` is idempotent

## Motivation and Context

Bugfix. `initializeDevices` doesn't work as expected

## Types of changes

### DIFF
--- a/examples/react-client/fishjam-chat/src/components/DevicePicker.tsx
+++ b/examples/react-client/fishjam-chat/src/components/DevicePicker.tsx
@@ -52,6 +52,22 @@ const DeviceSelect: FC<DeviceSelectProps> = ({ device }) => {
           Stream
         </Button>
       )}
+      <Button
+        disabled={!!device.stream}
+        onClick={async () => {
+          await device.initialize();
+        }}
+      >
+        Start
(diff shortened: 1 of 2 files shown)


czerwiukk (approved):

examples/react-client/fishjam-chat/src/components/DevicePicker.tsx
```diff
@@ -52,6 +52,22 @@ const DeviceSelect: FC<DeviceSelectProps> = ({ device }) => {
...
+      >
+        Start
+      </Button>
+      <Button
+        disabled={!device.stream}
+        onClick={async () => {
+          await device.stop();
+        }}
+      >
+        Stop
```
mironiasty, line 69: I know it is example app - but can we make these function more descriptive? Right now UI could have two stop buttons at some point.

kamil-stasiak: I've changed it to `Stop / Start streaming` and `Stop / Start device`

mironiasty (approved):

PR #91, opened 2024-09-05
FCE-498 Refactor start device function

kamil-stasiak:
## Description

If you invoke the `start` function without any parameters, it will start the last used device.

## Motivation and Context

It is very confusing to pass `true` as a `deviceId` to the `start` function in order to start the last used device.

## Types of Changes

Refactor


packages/react-client/src/DeviceManager.ts
```diff
@@ -130,11 +130,12 @@ export class DeviceManager
     this.storageConfig?.saveLastDevice(info);
   }
 
-  // todo `audioDeviceId / videoDeviceId === true` means use last device
-  public async start(deviceId?: string | boolean) {
-    const shouldRestart = !!deviceId && deviceId !== this.deviceState.media?.deviceInfo?.deviceId;
+  public async start(deviceId?: string) {
```
czerwiukk, line 133: **suggestion:** 
`const useLastDevice = deviceId === undefined;`
In this case `(deviceId || useLastDevice)` is almost always true, unless we pass as an empty string as an argument. (what is rather unwanted i guess)
```
  public async start(deviceId?: string) {
    const shouldRestart = !this.deviceState.media?.deviceInfo?.deviceId;
    const newDevice = deviceId ?? this.getLastDevice()?.deviceId ?? true;
```

packages/react-client/src/DeviceManager.ts
```diff
@@ -130,11 +130,12 @@ export class DeviceManager
...
+    const useLastDevice = deviceId === undefined;
+    const isDeviceStopped = this.deviceState.media?.deviceInfo?.deviceId === undefined;
+    const shouldRestart = (deviceId || useLastDevice) && isDeviceStopped;
 
-    const newDevice = deviceId === true ? this.getLastDevice()?.deviceId || true : deviceId;
+    const newDevice = useLastDevice ? this.getLastDevice()?.deviceId || true : deviceId;
```
mironiasty, line 138: Why `true` is used here?

packages/react-client/src/DeviceManager.ts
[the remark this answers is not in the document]
kamil-stasiak: If the user does not have a preference for which device to receive, they can start the last selected device (if possible) by invoking `start()`.

If the user wants to start a specific device, they should invoke this function with the device ID: `start(someId)`.

kamil-stasiak:
@mironiasty and @czerwiukk, have a look at the new implementation. The previous one had errors.

packages/react-client/src/DeviceManager.ts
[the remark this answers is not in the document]
czerwiukk: I think simplifying the invocation from `start(undefined)` to `start()` makes sense.

packages/react-client/src/DeviceManager.ts
```diff
@@ -130,22 +130,24 @@ export class DeviceManager
...
-    const exactConstraints = shouldRestart && prepareMediaTrackConstraints(newDevice, trackConstraints);
+    const exactConstraints = shouldProceed && prepareMediaTrackConstraints(newDeviceId, trackConstraints);
     if (!exactConstraints) return;
 
     this.deviceState.mediaStatus = "Requesting";
 
     this.emit(
       "devicesStarted",
-      { ...this.deviceState, restarting: shouldRestart, constraints: newDevice },
+      { ...this.deviceState, restarting: shouldReplaceDevice, constraints: newDeviceId },
```
mironiasty, line 150: Why we are passing `deviceId` as `constraints`?

packages/react-client/src/DeviceManager.ts
```diff
@@ -130,22 +130,24 @@ export class DeviceManager
...
-    const newDevice = deviceId === true ? this.getLastDevice()?.deviceId || true : deviceId;
+    const shouldReplaceDevice = Boolean(currentDeviceId && currentDeviceId !== newDeviceId);
+    const isDeviceStopped = currentDeviceId === undefined;
+    const shouldProceed = isDeviceStopped || shouldReplaceDevice;
 
     const trackConstraints = this.constraints;
 
-    const exactConstraints = shouldRestart && prepareMediaTrackConstraints(newDevice, trackConstraints);
+    const exactConstraints = shouldProceed && prepareMediaTrackConstraints(newDeviceId, trackConstraints);
     if (!exactConstraints) return;
```
mironiasty, line 144: I would consider refactoring it to separate function. `deviceId` being `true` for some random reason is... this just should not be written like that.

packages/react-client/src/DeviceManager.ts
kamil-stasiak: Good catch, fixed!

packages/react-client/src/DeviceManager.ts
kamil-stasiak: You're right, fixed!

mironiasty (approved):

PR #95, opened 2024-09-10
FCE 410 Add toggle function

kamil-stasiak:
## Description

The toggle function works in two modes: "soft" and "hard" (default). The user can invoke this function without any arguments. It starts (or resumes) the device and streams media if the user is connected. 

### `hard` 
- Turns off the device
- Slower start
- Turns off the device LED
- Example usage: camera

### `soft`
- Only disables the device
- Faster start
- The device LED remains active
- Example usage: microphone

## Motivation and Context

To unify the API with the Mobile SDK. It's important to implement this feature to avoid confusion:
https://linear.app/swmansion/issue/FCE-489/add-the-camera-and-mic-after-connecting-if-they-are-active-web

## Types of changes

New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/react-client/src/hooks/usePeerStatus.ts
+++ b/packages/react-client/src/hooks/usePeerStatus.ts
@@ -7,10 +7,13 @@ export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>
   const [peerStatus, setPeerStatus] = useState<PeerStatus>(null);
   const peerStatusRef = useRef<PeerStatus>(null);
 
-  const setInnerStatus = useCallback((status: PeerStatus) => {
-    peerStatusRef.current = status
-    setPeerStatus(status)
-  }, [setPeerStatus]);
+  const setInnerStatus = useCallback(
+    (status: PeerStatus) => {
+      peerStatusRef.current = status;
+      setPeerStatus(status);
+    },
+    [setPeerStatus],
+  );
 
   const getCurrentPeerState = useCallback(() => peerStatusRef.current, []);
 
--- a/packages/react-client/src/hooks/usePeerStatus.ts
+++ b/packages/react-client/src/hooks/usePeerStatus.ts
@@ -1,27 +1,40 @@
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import type { PeerStatus } from "../state.types";
 import type { FishjamClient } from "@fishjam-cloud/ts-client";
 import type { PeerMetadata, TrackMetadata } from "../types";
 
 export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>) => {
   const [peerStatus, setPeerStatus] = useState<PeerStatus>(null);
+  const peerStatusRef = useRef<PeerStatus>(null);
+
+  const setInnerStatus = useCallback((status: PeerStatus) => {
+    peerStatusRef.current = status
+    setPeerStatus(status)
+  }, [setPeerStatus]);
+
+  const getCurrentPeerState = useCallback(() => peerStatusRef.current, []);
 
   useEffect(() => {
+    const setConnecting = () => {
+      setInnerStatus("connecting");
+    };
     const setAuthenticated = () => {
-      setPeerStatus("authenticated");
+      setInnerStatus("authenticated");
     };
     const setError = () => {
-      setPeerStatus("error");
+      setInnerStatus("error");
     };
     const setJoined = () => {
-      setPeerStatus("joined");
+      setInnerStatus("joined");
     };
     const setDisconnected = () => {
-      setPeerStatus(null);
+      setInnerStatus(null);
     };
     const setConnected = () => {
-      setPeerStatus("connected");
+      setInnerStatus("connected");
     };
+
+    client.on("connectionStarted", setConnecting);
     client.on("authSuccess", setAuthenticated);
     client.on("joined", setJoined);
     client.on("authError", setError);
@@ -31,6 +44,7 @@ export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>
     client.on("socketOpen", setConnected);
 
     return () => {
+      client.off("connectionStarted", setConnecting);
       client.off("authSuccess", setAuthenticated);
       client.off("joined", setJoined);
       client.off("authError", setError);
@@ -39,7 +53,7 @@ export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>
       client.off("disconnected", setDisconnected);
       client.off("socketOpen", setConnected);
     };
-  }, [client]);
+  }, [client, setInnerStatus]);
 
-  return [peerStatus, setPeerStatus] as const;
+  return { peerStatus, getCurrentPeerState } as const;
 };
--- a/packages/react-client/src/hooks/usePeerStatus.ts
+++ b/packages/react-client/src/hooks/usePeerStatus.ts
@@ -4,37 +4,37 @@ import type { FishjamClient } from "@fishjam-cloud/ts-client";
 import type { PeerMetadata, TrackMetadata } from "../types";
 
 export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>) => {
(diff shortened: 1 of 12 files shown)


packages/react-client/src/types.ts
```diff
@@ -178,6 +178,8 @@ export type Device = {
 
 export type AudioDevice = Device & { isAudioPlaying: boolean };
 
+export type ToggleMode = "soft" | "hard";
```
kamil-stasiak, line 181: I believe we could come up with a better name. Here are some other suggestions:

1. Just a boolean: `toggle(forceStop: boolean = true)` 
2. `toggle(mode: "suspend" | "turnOff" = "turnOff")`
3. `toggle(onStop: "disable" | "stop" = "stop")`

However, I’m not satisfied with any of these options. Please help!

packages/react-client/src/hooks/useTrackManager.ts
```diff
@@ -18,15 +18,24 @@ export const useTrackManager = ({ mediaManager, tsClient }: TrackManagerConfig):
   const [paused, setPaused] = useState<boolean>(false);
   const [currentTrackMiddleware, setCurrentTrackMiddleware] = useState<TrackMiddleware>(null);
   const type = TRACK_TYPE_TO_DEVICE[mediaManager.getDeviceType()];
+  const joined = useRef<boolean>(false);
```
czerwiukk, line 21: **suggestion:** i think the `usePeerStatus` hook would've come in handy here

packages/react-client/src/hooks/useTrackManager.ts
```diff
@@ -132,6 +141,42 @@ export const useTrackManager = ({ mediaManager, tsClient }: TrackManagerConfig):
...
+        await resumeStreaming();
+      } else {
+        await startStreaming();
+      }
+    }
+  };
+
+  async function toggle(mode: ToggleMode = "hard") {
+    const mediaStream = mediaManager.getMedia()?.stream;
+    const track =
```
czerwiukk, line 156: **suggestion:** `mediaManager.getTracks()[0]`

packages/react-client/src/types.ts
czerwiukk: i am more of a fan of implementing two functions like
`toggleDevice`
and 
`toggleStreaming`.
we get similar logic and we don't need to pass an argument so the toggle function can be put as an event handler

packages/react-client/src/hooks/useTrackManager.ts
kamil-stasiak: Unfortunately, the status from `usePeerStatus` could be outdated.

If I understand the React flow correctly, the toggle function remembers the status of the invocation. So, if it starts with null, it's possible that this state changes to joined while the device is starting, but the stream functions will remember the outdated state. Additionally, the toggle function will still have a reference to the outdated stream method.

czerwiukk: Every function is also recomputed at react render. So if you use a state and capture it in a closure, it won't go out of sync, because after a render, that's a different function with different closure state.

czerwiukk: https://codesandbox.io/p/sandbox/mt4l88

compare calling those two `setDisplayedMessage` variants, they will behave the same

packages/react-client/src/hooks/useTrackManager.ts
kamil-stasiak: image
image
On my machine, the first one is always `audio`

czerwiukk: please have another look, i wrote `mediaManager.getTracks()`, not `mediaStream.getTracks()`

packages/react-client/src/types.ts
kamil-stasiak: I like that approach; it has a few advantages as well as disadvantages:

## Advantages
- Greater granularity
- Predictable behaviour
- plus everything you just mentioned

## Disadvantages
- Two functions instead of one; the user has to invoke both of them.
- On mobile, we don't have such separation; maybe we should add them there.
- To be compatible with the approach that I proposed, we need to either disable the stream in `toggleStreaming` to achieve an immediate `mute` effect (which is strange, because we just stopped streaming) or add separate `toggleMute / toggleDisable` functions, which would introduce another `toggle` function 😕

So for now, I want to merge this single powerful function

packages/react-client/src/hooks/useTrackManager.ts
kamil-stasiak: image

kamil-stasiak: image

packages/react-client/src/hooks/useTrackManager.ts
czerwiukk: I'm pushing this solution because I want to keep the code here simple.
If we need a ref with `joined` state - it should land in another hook like `useJoinedStatusRef`.
Though I think we don't need that and already created hook should make do here.

kamil-stasiak: > `joined.current` is used at the top of the function before any async code, so we can be certain the value hasn't changed between invoking the function and using the value.

Yeah, this value is stable state inside this function.

The problem arises when the user invokes the toggle function. That function saves a reference to the `stream` function, which holds a particular state at the time it was called. Even if the `peerState` changes, it will not be reflected in future `toggle` invocations.

```
const peerStatus = usePeerStatus()
const stream = async () => {
    // peerStatus at the moment of `toggle` invocation 
    if (!joined.current) return; // it does not deopend on memoized value because it mutates

    if (currentTrack?.trackId) {
      await resumeStreaming();
    } else {
      await startStreaming();
    }
 };

  async function toggle(mode: ToggleMode) {
    // ... 
      await mediaManager.start(true); // async
      // peerState can chage here
      await stream(); // reference to function with outdated state, function with new state will be available in next invocation  
  }
```

czerwiukk: alright, let's add `getCurrentPeerState` method based on a ref in `usePeerState`

packages/react-client/src/hooks/usePeerStatus.ts
```diff
@@ -1,27 +1,40 @@
...
 import type { PeerMetadata, TrackMetadata } from "../types";
 
 export const usePeerStatus = (client: FishjamClient<PeerMetadata, TrackMetadata>) => {
   const [peerStatus, setPeerStatus] = useState<PeerStatus>(null);
+  const peerStatusRef = useRef<PeerStatus>(null);
+
+  const setInnerStatus = useCallback((status: PeerStatus) => {
+    peerStatusRef.current = status
+    setPeerStatus(status)
+  }, [setPeerStatus]);
```
czerwiukk, line 13: **nitpick (non-blocking):** 
```suggestion
  const [peerStatus, setPeerStatusState] = useState<PeerStatus>(null);
  const peerStatusRef = useRef<PeerStatus>(null);

  const setPeerStatus = useCallback((status: PeerStatus) => {
    peerStatusRef.current = status
    setPeerStatusState(status)
  }, [setPeerStatusState]);
```

czerwiukk: invocations are also required to be changed smh

mironiasty (approved):

packages/react-client/src/types.ts
```diff
@@ -197,6 +199,23 @@ export interface TrackManager {
...
+   * - **Hard Mode** - Turns the physical device on and off.
+   *   - If started: disables the media stream, pauses streaming, and stops the device.
+   *   - If stopped: starts the device and begins (or resumes) streaming.
+   *
+   * - **Soft Mode** - Enables and disables the media stream. Starts the device if needed.
+   *   - If enabled: disables the media stream and pauses streaming, but does not stop the device.
+   *   - If disabled: enables the media stream and starts (or resumes) streaming.
+   *   - If stopped: starts the device, enables the media stream, and starts (or resumes) streaming.
+   */
+  toggle: (mode?: ToggleMode) => Promise<void>;
```
kamil-stasiak, line 218: @mironiasty WDYT?

PR #97, opened 2024-09-11
FCE-460 Return a promise from the connect function

kamil-stasiak:
## Description

The connect function now returns a promise instead of a "disconnect" callback.

## Motivation and Context

This change allows users to implement custom logic upon connection instead of using the `joined` event.

## Types of changes

### DIFF
--- a/packages/ts-client/src/ConnectPromise.ts
+++ b/packages/ts-client/src/ConnectPromise.ts
@@ -1,0 +1,40 @@
+import { Deferred } from './webrtc/deferred';
+import type { FishjamClient } from './FishjamClient';
+
+export class ConnectPromise<P, T> {
+  private readonly result: Deferred<void>;
+  private readonly clearCallbacks: () => void;
+
+  constructor(fishjamClient: FishjamClient<P, T>) {
+    this.result = new Deferred<void>();
+
+    const onSuccess = () => {
+      this.result.resolve();
+    };
+    const onError = () => {
+      this.result.reject('joinError');
+    };
+
+    fishjamClient.on('joined', onSuccess);
+    fishjamClient.on('joinError', onError);
+    fishjamClient.on('authError', onError);
+    fishjamClient.on('socketError', onError);
+
+    this.clearCallbacks = () => {
+      fishjamClient.removeListener('joined', onSuccess);
+      fishjamClient.removeListener('joinError', onError);
+      fishjamClient.removeListener('authError', onError);
+      fishjamClient.removeListener('socketError', onError);
+    };
+  }
+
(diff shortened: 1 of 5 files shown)


czerwiukk (approved):

packages/ts-client/src/ConnectPromise.ts
```diff
@@ -0,0 +1,40 @@
+import { Deferred } from './webrtc/deferred';
+import type { FishjamClient } from './FishjamClient';
+
+export class ConnectPromise<P, T> {
```
mironiasty, line 4: WDYT about using just promise instead of full new class?
```ts
export function connectPromiseFun<T, P>(fishjamClient: FishjamClient<P, T>) {
  const promise = new Promise<void>((resolve, reject) => {
    const onSuccess = () => {
      clearCallbacks();
      resolve();
    };
    const onError = () => {
      clearCallbacks();
      reject();
    };

    fishjamClient.on('joined', onSuccess);
    fishjamClient.on('joinError', onError);
    fishjamClient.on('authError', onError);
    fishjamClient.on('socketError', onError);

    const clearCallbacks = () => {
      fishjamClient.removeListener('joined', onSuccess);
      fishjamClient.removeListener('joinError', onError);
      fishjamClient.removeListener('authError', onError);
      fishjamClient.removeListener('socketError', onError);
    };
  });

  return promise;
}
```

> [!NOTE]
> Name `connectPromiseFun` is just example, could be something more descriptive - like `connectEventsHandler`?

packages/ts-client/src/FishjamClient.ts
```diff
@@ -314,11 +315,15 @@ export class FishjamClient<PeerMetadata, TrackMetadata> extends (EventEmitter as
...
-  connect(config: ConnectConfig<PeerMetadata>): void {
+  public async connect(config: ConnectConfig<PeerMetadata>): Promise<void> {
+    const result = new ConnectPromise(this);
+
     this.reconnectManager.reset(config.peerMetadata);
     this.connectConfig = config;
 
     this.initConnection(config.peerMetadata);
+
+    return result.getPromise();
```
mironiasty, line 326: with above change, you can also simplify calls here:
```ts
    const result = connectPromiseFun(this);
```
and
```suggestion
    return result;
```
(or `return await result` to be more explicit)

packages/ts-client/src/ConnectPromise.ts
kamil-stasiak: It's nice and elegant. I like it!

mironiasty (approved):
Bear Lets Merge

PR #102, opened 2024-09-17
FCE-543 Remove warn on every disconnect

kamil-stasiak:
## Description

`stopStreaming()` on disconnect is only invoked if there is a screen share stream.

## Motivation and Context

There is no need to issue a warning about screen sharing on every disconnect.

## Types of changes

### DIFF
--- a/packages/react-client/src/hooks/useScreenShare.ts
+++ b/packages/react-client/src/hooks/useScreenShare.ts
@@ -20,6 +20,9 @@ export const useScreenShare = (): ScreenshareApi => {
 
   const tsClient = fishjamClientRef.current;
 
+  const stream = state.stream ?? null;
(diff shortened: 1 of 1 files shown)


packages/react-client/src/hooks/useScreenShare.ts
```diff
@@ -20,6 +20,9 @@ export const useScreenShare = (): ScreenshareApi => {
 
   const tsClient = fishjamClientRef.current;
 
+  const stream = state.stream ?? null;
+  const [videoTrack, audioTrack] = stream ? getTracks(stream) : [null, null];
```
mironiasty, line 24: note: we should add eslint rule about variable shadowing in this project...

mironiasty (approved):
LGTM

PR #103, opened 2024-09-18
FCE-458 Rename client statuses

kamil-stasiak:
## Description

List of peers statuses has been shortened to: `connecting`, `connected`, `error`, `idle`.

Removed: `connected`, `authenticated`, `closed`, `null`;
Renamed: `joined` -> `connected`

## Motivation and Context

- The status `joined` was not descriptive.
- The user probably doesn't care about intermediate statuses (authenticated, web socket closed, web socket connected).

## Types of changes

### DIFF
--- a/packages/react-client/src/state.types.ts
+++ b/packages/react-client/src/state.types.ts
@@ -36,4 +36,12 @@ export type PeerState = {
   tracks: Record<TrackId, Track>;
 };
 
-export type PeerStatus = "connecting" | "connected" | "error" | "idle";
+/**
+ * Represents the possible statuses of a peer connection.
+ *
+ * idle - Participant is not connected, either never connected or successfully disconnected.
+ * connecting - Participant is in the process of connecting.
+ * connected - Participant has successfully connected.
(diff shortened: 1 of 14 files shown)


packages/react-client/src/state.types.ts
```diff
@@ -36,4 +36,4 @@ export type PeerState = {
   tracks: Record<TrackId, Track>;
 };
 
-export type PeerStatus = "connecting" | "connected" | "authenticated" | "joined" | "error" | "closed" | null;
+export type PeerStatus = "connecting" | "connected" | "error" | "idle";
```
kamil-stasiak, line 39: Should I change `PeerStatus` to `ParticipantStatus` or `ConnectionStatus`?

mironiasty: No strong opinion here 🤔 Both seems legit.

mironiasty: btw, can we add some comment/other info about difference between connected and idle?

mironiasty (approved):

PR #104, opened 2024-09-18
FCE-549 Allow only one track of each kind

kamil-stasiak:
## Description

`useParticipants()` returns a single track instead of an array because streaming more than one track shouldn't be possible.

Every instance of `screenshare` has been renamed to `screenShare`.

## Motivation and Context

A participant can only stream one track of each kind, so the `useParticipant` hook reflects this behaviour.

## Types of changes


examples/react-client/use-camera-and-microphone/src/MainControls.tsx
```diff
@@ -87,12 +87,10 @@ export const MainControls = () => {
...
-  const localTracks = localParticipant
-    ? [
-        ...localParticipant.cameraTracks,
-        ...localParticipant.screenshareVideoTracks,
-      ]
-    : [];
+  const localTracks = [
+    localParticipant?.cameraTrack,
+    localParticipant?.screenShareVideoTrack,
+  ];
```
czerwiukk, line 93: **suggestion:** instead of putting `?.fiter` in line 394, we could do sth like that here.
```suggestion
  const localTracks = [
    localParticipant?.cameraTrack,
    localParticipant?.screenShareVideoTrack,
  ].filter((track): track is Track => Boolean(track));
```

czerwiukk (approved):
just one suggestion

PR #107, opened 2024-09-19
FCE-554

kamil-stasiak:
### The old `useFishjamClient_DO_NOT_USE`
- Methods from `useFishjamClient_DO_NOT_USE` like `connnect`, `addTrack` etc. were deleted because they weren't used.
- `useFishjamClient_DO_NOT_USE` was renamed to `useFishjamClientState` because it only holds state.
- There is no need to expose this state, so it isn't exported.
- There is only one invocation of this hook in `FishjamProvider` so there is only one active subscription. In the previous implementation, `useParticipants` creates a new subscription (`useSyncExternalStore`).

### The new `useFishjamClient_DO_NOT_USE`

To get access to methods and events from `ts-client` a new hook was introduced: `useFishjamClient_DO_NOT_USE`.

## Motivation and Context

`useFishjamClient_DO_NOT_USE` didn't exposed methods like `getStatistics()` and `events`

## Types of changes

### DIFF
--- a/packages/react-client/src/hooks/internal.ts
+++ b/packages/react-client/src/hooks/internal.ts
@@ -1,1 +1,1 @@
-export { useFishjamClient_DO_NOT_USE } from "./useFishjamClient";
+export { useFishjamClient_DO_NOT_USE } from "./useFishjamClient_DO_NOT_USE";
--- a/packages/react-client/src/fishjamProvider.tsx
+++ b/packages/react-client/src/fishjamProvider.tsx
@@ -5,6 +5,7 @@ import { FishjamClient, type ReconnectConfig } from "@fishjam-cloud/ts-client";
 import { FishjamContext } from "./hooks/useFishjamContext";
 import { DeviceManager } from "./DeviceManager";
 import { usePeerStatus } from "./hooks/usePeerStatus";
+import { useFishjamClientState } from "./hooks/useFishjamClientState";
 
 interface FishjamProviderProps extends PropsWithChildren {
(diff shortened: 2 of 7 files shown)


packages/react-client/src/hooks/internal.ts
```diff
@@ -1 +1 @@
-export { useFishjamClient_DO_NOT_USE } from "./useFishjamClient";
+export { useFishjamClient_DO_NOT_USE } from "./useFishjamClient_DO_NOT_USE";
```
mironiasty, line 1: Why we need this file? 🤔

mironiasty (approved):

packages/react-client/src/hooks/internal.ts
kamil-stasiak: For the Videoroom to continue working. The Videoroom logic requires events and methods from the `ts-client`. With current API it's impossible to e.g. request different simulcast layer

czerwiukk: i think we should extend our api to cover videoroom needs in the future. lets do it like that rn but leave a followup ticket

kamil-stasiak: I totally agree; the videoroom needs to be simplified, and our new API should be flexible enough to support that.
https://linear.app/swmansion/issue/FCE-560/simplify-the-videroom-codebase
https://linear.app/swmansion/issue/FCE-561/remove-usefishjamclient-do-not-use

PR #109, opened 2024-09-20
FCE-562 Restore participant metadata

kamil-stasiak:
## Description

`metadata` field now holds `peer` and `server` metadata. 

SDK doesn't handle server metadata at all. There is a separate task for that: 
https://linear.app/swmansion/issue/FCE-563/handle-server-metadata-in-web-client-sdk 

This is a basic fix. I didn't want to refactor or redesign anything.

## Motivation and Context

Peer / Participant metadata didn't work.

## Types of changes


packages/react-client/src/state.types.ts
```diff
@@ -30,7 +30,10 @@ export type TrackWithOrigin = Track & {
 
 export type PeerState = {
   id: PeerId;
-  metadata?: PeerMetadata;
+  metadata: {
+    peer?: PeerMetadata;
+    server: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
czerwiukk, line 35: **suggestion:** 
```suggestion
    server: Record<string, unknown>
```

czerwiukk (approved):

PR #118, opened 2024-09-26
FCE-583 Back at it again with the `peer` word

czerwiukk:
## Description

Bring back p word from another p word.

## Motivation and Context

It has turned out to be a simpler approach.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/types.ts
+++ b/packages/react-client/src/types.ts
@@ -267,8 +267,8 @@ type DistinguishedTracks = {
 export type PeerStateWithTracks = PeerState & DistinguishedTracks;
 
 export type Participiants = {
-  localParticipant: PeerStateWithTracks | null;
-  participants: PeerStateWithTracks[];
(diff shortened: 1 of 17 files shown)


packages/react-client/src/types.ts
```diff
@@ -267,8 +267,8 @@ type DistinguishedTracks = {
 export type PeerStateWithTracks = PeerState & DistinguishedTracks;
 
 export type Participiants = {
```
kamil-stasiak, line 269: I found another one with typo

mironiasty: Is this type even required?

mironiasty (approved):
There is one comment from Kamil
Rest looks good 👍

packages/react-client/src/types.ts
czerwiukk: not used - i removed it

PR #119, opened 2024-09-27
FCE-552 Add constraints to getUserMedia

kamil-stasiak:
#### This configuration remains optional.
`<FishjamProvider>`  
Default constraints will be used if no configuration is provided.

#### Users can override default values.  
`<FishjamProvider deviceManagerConfig={{ videoTrackConstraints: { frameRate: 30 } }}>`  
User configurations will be merged with the default constraints.

#### Users can disable audio and/or video.  
`<FishjamProvider deviceManagerConfig={{ videoTrackConstraints: false, audioTrackConstraints: false }}>`

## Motivation and Context

To set consistent constraints and provide an option to override them.

## Types of Changes


packages/react-client/src/fishjamProvider.tsx
```diff
@@ -1,27 +1,45 @@
...
 import { FishjamContext } from "./hooks/useFishjamContext";
 import { DeviceManager } from "./DeviceManager";
 import { useParticipantStatus } from "./hooks/useParticipantStatus";
 import { useFishjamClientState } from "./hooks/useFishjamClientState";
+import { AUDIO_TRACK_CONSTRAINTS, VIDEO_TRACK_CONSTRAINTS } from "./constraints";
 
 interface FishjamProviderProps extends PropsWithChildren {
   config?: { reconnect?: ReconnectConfig | boolean };
-  deviceManagerDefaultConfig?: DeviceManagerConfig;
+  deviceManagerConfig?: DeviceManagerUserConfig;
```
czerwiukk, line 14: **suggestion:** user doesn't know anything about a device manager. the property could be just named `defaultConstraints` or `constraints`

packages/react-client/src/constraints.ts
```diff
@@ -28,28 +28,37 @@ export const SCREEN_SHARING_MEDIA_CONSTRAINTS: MediaStreamConstraints = {
...
-  }
-  return constraint;
+  defaultConstraints: MediaTrackConstraints,
+  userConstraints?: boolean | MediaTrackConstraints,
+): boolean | MediaTrackConstraints => {
+  if (userConstraints === false) return false;
+
+  if (!userConstraints) return defaultConstraints;
+
+  return userConstraints;
```
czerwiukk, line 38: **nitpick:** 
the whole function could be replaced with this expression
```
userConstraints ?? defaultConstraints
```

packages/react-client/src/constraints.ts
```diff
@@ -28,28 +28,37 @@ export const SCREEN_SHARING_MEDIA_CONSTRAINTS: MediaStreamConstraints = {
...
 export const prepareConstraints = (
   deviceIdToStart: string | undefined,
-  constraints: MediaTrackConstraints | undefined,
-): MediaTrackConstraints | undefined => {
+  constraints: MediaTrackConstraints | undefined | boolean,
+): MediaTrackConstraints | undefined | boolean => {
   if (!deviceIdToStart) return constraints;
 
-  return { ...constraints, deviceId: { ideal: deviceIdToStart } };
+  // does not start this device
```
czerwiukk, line 58: **question:** what does it mean?

packages/react-client/src/fishjamProvider.tsx
czerwiukk: storage, if actually needed, could be an another property such as persistence options or sth

packages/react-client/src/constraints.ts
kamil-stasiak: I rephrased it to:
```
// The resulting stream will not contain a track of this type,
// which means that the device will not be activated.
```

packages/react-client/src/fishjamProvider.tsx
kamil-stasiak: WDYT about it?

```
interface FishjamProviderProps extends PropsWithChildren {
  config?: { reconnect?: ReconnectConfig | boolean };
  constraints?: Pick<MediaStreamConstraints, "audio" | "video">;
  persistLastDevice?: boolean | PersistLastDeviceHandlers;
}
```

czerwiukk: please unpack reconnect from config and i'm satisfied 💃

kamil-stasiak: updated ✅

czerwiukk (approved):

PR #120, opened 2024-09-27
FCE-581 tidy types

czerwiukk:
## Description

Remove excess types and group them by public/internal.

## Types of changes

      not work as expected)


packages/react-client/src/index.ts
```diff
@@ -13,24 +13,7 @@ export {
...
-  UseSetupMediaConfig,
-  ScreenshareApi,
-  UseConnect,
-  ConnectConfig,
-  TrackMiddleware,
-  PeerStateWithTracks,
-  TrackMetadata, // only for compatibility reasons, will be removed in: FCE-415
-  PeerMetadata, // only for compatibility reasons, will be removed in: FCE-415
-} from "./types";
+export * from "./types/public";
```
mironiasty, line 16: I'm not sure if this approach is better than old one.
With old one approach, you need to add type to export list explicit. Right now, any type added to public.ts will be 'public'. And it might happen by accident, as people tend to not read exact file names sometimes.

czerwiukk: we might as well do it explicitly

kamil-stasiak (commented):
LGTM 👍 

During the code review, I found a couple of other issues not related to your changes. Please consider whether it is worth fixing them

1. `useScreenShare` returns `ScreenShareApi`, but it is located in `internal.ts` and is not exported in `index.ts`.
2. Typo: `ScreenshareApi` -> `ScreenShareApi`.
3. `MediaDeviceType`, `TrackType`, `Devices`, and `UserMediaAPI` are never used.
4. Shouldn't `useVideoDeviceManager` and `useAudioDeviceManager` be removed from `poublic.ts`? Users should use `useCamera` and `useMicrophone` instead.

mironiasty (approved):

PR #124, opened 2024-10-01
Fix simple app example

roznawsk:
## Description

Local storage caused the same token to be set for all browser tabs. This resulted client B connecting, kicking out the peer A, then peer A tried to reconnect, and so on.

## Motivation and Context

## Types of changes

      not work as expected)


czerwiukk (approved):
image

PR #126, opened 2024-10-03
FCE-650 Fix replace track promise

kamil-stasiak:
## Description

The command handler waits for the `replaceTrack` promise to resolve.

## Motivation and Context

If there were two `replaceTrack` calls in a row, the second one would not resolve until a new command entered the `commandsQueue`.

## Types of changes

### DIFF
--- a/packages/ts-client/src/webrtc/webRTCEndpoint.ts
+++ b/packages/ts-client/src/webrtc/webRTCEndpoint.ts
@@ -429,7 +429,7 @@ export class WebRTCEndpoint<EndpointMetadata = any, TrackMetadata = any> extends
       stream.addTrack(track);
 
       this.commandsQueue.pushCommand({
-        handler: () => {
+        handler: async () => {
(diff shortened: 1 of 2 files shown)


packages/ts-client/src/webrtc/webRTCEndpoint.ts
```diff
@@ -574,7 +572,7 @@ export class WebRTCEndpoint<EndpointMetadata = any, TrackMetadata = any> extends
     const resolutionNotifier = new Deferred<void>();
 
     this.commandsQueue.pushCommand({
-      handler: () => {
+      handler: async () => {
         this.localTrackManager.removeTrackHandler(trackId);
       },
```
mironiasty, line 577:
```suggestion
      handler: async () => this.localTrackManager.removeTrackHandler(trackId),
```
So if removeTrackHandler will be async in future, it will work correctly

mironiasty (approved):

czerwiukk (approved):
approved after resolving Miron's comment

PR #128, opened 2024-10-03
FCE-626 middleware starts two times

kamil-stasiak:
## Description

Logic responsible for managing middleware was moved from `useTrackManager` to the new class `MiddlewareManager` which is internally used in `DeviceManager`. Device manager is reposible

## Refactor
- The `storageConfig` from `DeviceManager` has been extracted to `localStorage`.
- Some functions have been moved to different files to avoid unnecessary exports.
- `useProcessedPreviewStream.ts`, `useVideoDeviceManager.ts` and `useVideoDeviceManager.ts` have been removed.

## Motivation and Context

The middleware was not handled correctly. It was executed twice for each call

## Types of changes


czerwiukk (dismissed):

czerwiukk (changes requested):
missclicked earlier 😭

packages/react-client/src/devices/MiddlewareManager.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.media.track.enabled = false;
+    this.media.enabled = false;
+  }
+
+  public enable() {
+    if (!this.media?.track) return;
+
+    this.media!.track!.enabled = true;
+    this.media!.enabled = true;
+  }
```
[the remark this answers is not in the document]
kamil-stasiak: Ad 2. There are two distinct pairs of Stream + Track. One is the raw stream from a device, while the second is the result of the raw stream processed by middleware. Stopping the raw stream probably doesn't stop the middleware stream, because the latter is purely virtual, and the middleware is responsible for updating the stream

packages/react-client/src/hooks/useScreenShare.ts
```diff
@@ -152,3 +151,10 @@ export const useScreenShare = () => {
 
   return screenShareManager;
 };
+
+export const getTracksFromStream = (stream: MediaStream): [MediaStreamTrack, MediaStreamTrack | null] => {
+  const video = stream.getVideoTracks()[0];
+  const audio = stream.getAudioTracks()[0] ?? null;
+
+  return [video, audio];
+};
```
kamil-stasiak, line 160: Moved from `media.ts`

packages/react-client/src/devices/MiddlewareManager.ts
```diff
@@ -0,0 +1,58 @@
...
+    return this.applyMiddleware(track, this.middleware);
+  }
+
+  public setTrackMiddleware(
+    middleware: TrackMiddleware | null,
+    rawTrack: MediaStreamTrack | null,
+  ): MiddlewareMedia | null {
+    this.middleware = middleware;
+    return this.applyMiddleware(rawTrack, middleware);
+  }
```
kamil-stasiak, line 23: Controversial Decision: The `updateMedia` and `setTrackMiddleware` functions mutate the state and return a new object. While this violates the Command Query Separation principle, I believe it's acceptable in this context. If we decide not to return from these functions, we need to store the result of `applyMiddleware` and then expose a method that allows us to read this field.

packages/react-client/src/devices/MiddlewareManager.ts
```diff
@@ -0,0 +1,58 @@
...
+      async () => {
+        this.clearMiddleware();
+        track.stop();
+      },
+    );
+
+    return {
+      stream: new MediaStream([track]),
+      track,
+    };
```
czerwiukk, line 56: **thought:** isn't this structure redundant?

packages/react-client/src/devices/MiddlewareManager.ts
czerwiukk: Or just use `updateMedia` explicitly when setting a new middleware. It separates the purposes really well.

packages/react-client/src/devices/MiddlewareManager.ts
```diff
@@ -0,0 +1,58 @@
...
+export type MiddlewareMedia = {
+  stream: MediaStream;
+  track: MediaStreamTrack;
+};
+
+export class MiddlewareManager {
+  private middleware: TrackMiddleware | null = null;
+  private clearMiddlewareFn: (() => void) | null = null;
+
+  public updateMedia(track: MediaStreamTrack | null): MiddlewareMedia | null {
```
czerwiukk, line 13: **thought:** 
what u think?
```suggestion
  public getProcessedMedia(track: MediaStreamTrack | null): MiddlewareMedia | null {
```

packages/react-client/src/DeviceManager.ts
```diff
@@ -25,49 +38,61 @@ export type DeviceManagerEvents = {
...
-    error: null,
-  };
+  private readonly deviceType: DeviceType;
+  private readonly middlewareManager = new MiddlewareManager();
+
+  public readonly getLastDevice: () => MediaDeviceInfo | null = () => null;
+  private readonly saveLastDevice: (info: MediaDeviceInfo) => void = NOOP;
+
+  private rawMedia: Media | null = null;
+  private middlewareMedia: MiddlewareMedia | null = null;
```
czerwiukk, line 69: **thought:** 
what do you think?
```suggestion
  private processedMedia: MiddlewareMedia | null = null;
```

packages/react-client/src/DeviceManager.ts
```diff
@@ -25,49 +38,61 @@ export type DeviceManagerEvents = {
...
-    if (storage === true || storage === undefined) return getLocalStorageConfig(deviceType);
-    return storage;
+  public getState(): DeviceState {
+    return {
+      mediaStatus: this.mediaStatus,
+      devices: this.devices,
+      devicesStatus: this.devicesStatus,
+      error: this.error,
+      media: this.getMedia(),
+    };
```
czerwiukk, line 95: **suggestion:** put `currentMiddleware` here

packages/react-client/src/hooks/devices.ts
```diff
@@ -1,23 +1,22 @@
...
-  const { deviceState, status } = useVideoDeviceManager();
+  const { videoTrackManager, videoDeviceManagerRef } = useFishjamContext();
+  const { deviceState, status } = useDeviceManager(videoDeviceManagerRef.current);
+
   const { currentTrack, ...trackManager } = videoTrackManager;
-  const processedPreviewStream = useProcessedPreviewStream(videoTrackManager, deviceState.media?.track);
 
-  const stream = currentTrack?.stream ?? processedPreviewStream ?? deviceState.media?.stream ?? null;
+  const stream = deviceState.media?.stream ?? null;
+  const currentTrackMiddleware = videoDeviceManagerRef.current.getMiddleware();
```
czerwiukk, line 19: **suggestion:** current middleware should be added in device state

packages/react-client/src/devices/MiddlewareManager.ts
kamil-stasiak: Doing so `setTrackMiddleware` still violates the Command Query Separation. `applyMiddleware` mutates internal state

packages/react-client/src/devices/MiddlewareManager.ts
kamil-stasiak: What do you mean? I don't understand

packages/react-client/src/DeviceManager.ts
kamil-stasiak: Now it is `processedMediaTrack`

packages/react-client/src/devices/MiddlewareManager.ts
```diff
@@ -50,9 +44,6 @@ export class MiddlewareManager {
       },
     );
 
-    return {
-      stream: new MediaStream([track]),
-      track,
-    };
+    return track;
   }
 }
```
czerwiukk, line 49: **suggestion:** 
```
export class MiddlewareManager {
  private middleware: TrackMiddleware | null = null;
  private clearMiddlewareFn: (() => void) | null = null;

  public getMiddleware(): TrackMiddleware | null {
    return this.middleware;
  }

  public clearMiddleware() {
    this.clearMiddlewareFn?.();
    this.clearMiddlewareFn = null;
  }

  public applyMiddleware(
    rawTrack: MediaStreamTrack | null,
    middleware: TrackMiddleware | null = this.middleware,
  ): MediaStreamTrack | null {
    this.clearMiddleware();
    if (!rawTrack || !middleware) return null;
    const { onClear, track } = middleware(rawTrack);

    if (onClear) this.clearMiddlewareFn = onClear;

    setupOnEndedCallback(
      track,
      () => track.id,
      async () => {
        this.clearMiddleware();
        track.stop();
      },
    );

    return track;
  }
}
```

what do you think?

czerwiukk:
```
  public async setTrackMiddleware(middleware: TrackMiddleware | null): Promise<void> {
    const rawTrack = this.rawMedia?.track ?? null;
    const processedMedia = this.middlewareManager.applyMiddleware(rawTrack, middleware);
    this.setCurrentMedia(this.rawMedia, processedMedia);

    this.emit("middlewareSet", this.getState());
  }
```

```
  private updateMedia(media: Media | null) {
    this.rawMedia = !media ? null : { ...media };

    const processedMedia = this.middlewareManager.applyMiddleware(media?.track ?? null);
    this.setCurrentMedia(this.rawMedia, processedMedia);
  }
```

czerwiukk (approved):

PR #135, opened 2024-10-09
Expose isMuted and device error

czerwiukk:
## Description

Expose missing properties.

## Types of changes

      not work as expected)


kamil-stasiak (changes requested):
`isMuted` doesn't work because the internal state isn't being updated.

We need to add at least something like this to `DeviceManager` to fix it: 

```ts
private setEnable = (track: MediaStreamTrack | null, value: boolean) => {
    if (!track) return;
    track!.enabled = value;

    if (this.media) {
      this.media.enabled = value // that's missing
    }
  };
```

kamil-stasiak (approved):

PR #143, opened 2024-10-16
FCE-720 VAD hook

czerwiukk:
## Description

Implements `useVAD` hook allowing users to listen to VAD updates of selected peers.
Introduces changes to the way `fishjamClientState` stores peers data.
The `TrackContext` is kept as it is in the `FIshjamClient` to allow access to the track events.
Transformations of the peers now happens in the `usePeers` hook, just before exposing the data outside.

## Motivation and Context

Users should have an opt-in way to listen to VAD updates.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/hooks/useFishjamClientState.ts
+++ b/packages/react-client/src/hooks/useFishjamClientState.ts
@@ -1,7 +1,6 @@
 import { useCallback, useMemo, useRef, useSyncExternalStore } from "react";
-import type { Component, Endpoint, MessageEvents, Peer, TrackContext, FishjamClient } from "@fishjam-cloud/ts-client";
-import type { PeerMetadata, PeerState, TrackId, TrackMetadata } from "../types/internal";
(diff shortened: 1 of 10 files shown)


packages/react-client/src/hooks/useFishjamClientState.ts
```diff
@@ -44,41 +43,12 @@ const eventNames = [
 ] as const satisfies (keyof MessageEvents<unknown, unknown>)[];
 
 export interface FishjamClientState {
-  peers: PeerState[];
-  components: PeerState[];
-  localPeer: PeerState | null;
+  peers: Record<string, Peer<PeerMetadata, TrackMetadata>>;
```
kamil-stasiak, line 46: Nit: There is an alias for the `string` type in `types/interna.ts` that indicates this particular string is a `PeerId`

kamil-stasiak (approved):

PR #145, opened 2024-10-17
FCE-727 Nullify streams

kamil-stasiak:
## Description

The `stream` field exposes only `deviceState.media?.stream`.

## Motivation and Context

The camera and microphone streams were not nullified after turning off the device because a stream from a remote connection was exposed.

## Types of changes


examples/react-client/fishjam-chat/src/components/Tile.tsx
```diff
@@ -24,7 +24,9 @@ export function Tile({ videoTrack, audioTrack, name, id }: Props) {
       )}
 
       <div className="absolute bottom-2 left-0 z-30 grid w-full place-content-center text-center text-xs">
-        <AudioVisualizer stream={audioTrack?.stream} />
+        {/* The MediaStream from the local peer may not contain an audio track */}
+        {/* if the user has paused their stream  */}
+        {audioTrack?.track && <AudioVisualizer stream={audioTrack?.stream} />}
```
czerwiukk, line 29: **question:** what if we handled lack of a track in the visualizer by displaying nothing?

czerwiukk: alright - pass in the track as the prop and it'll work well

czerwiukk (approved):

PR #157, opened 2024-10-22
FCE-759 Hide the error if the user closes the screen share window

kamil-stasiak:
## Description

When the screen share window is closed, `NotAllowedError` is thrown. The `Share the screen` handler swallows this error.

## Motivation and Context

When the user changes their mind during screen sharing and closes the browser window, an error appears in the console. This is a completely normal situation, so we should not log this error to the console.

## Types of changes

### DIFF
--- a/examples/react-client/fishjam-chat/src/components/CallToolbar.tsx
+++ b/examples/react-client/fishjam-chat/src/components/CallToolbar.tsx
@@ -36,6 +36,16 @@ export const CallToolbar = () => {
   const CameraIcon = cameraStream ? Video : VideoOff;
   const ScreenshareIcon = screenStream ? MonitorOff : MonitorUp;
 
+  const toggleScreenShare = async () => {
+    if (screenStream) return stopStreaming();
+    try {
+      await startStreaming();
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


examples/react-client/fishjam-chat/src/components/DevicePicker.tsx
```diff
@@ -129,7 +129,20 @@ export function DevicePicker() {
...
+                if (
+                  error instanceof Error &&
+                  error.name === "NotAllowedError"
+                ) {
+                  // The user closed the screen share window
+                } else {
+                  console.error(error);
+                }
+              }
+            }}
```
czerwiukk, line 145: **nitpick:** please extract the function, it's grown enough
```
try {
    await screenShare.startStreaming();
  } catch (error) {
    if (
      error instanceof Error &&
      error.name === "NotAllowedError"
    ) return;
    console.error(error);
  }
```

czerwiukk: btw, is `NotAllowedError` only being raised when user close the screenshare window?

czerwiukk (approved):

PR #158, opened 2024-10-22
Remove peer/server structure

roznawsk:
## Description

Remove the requirement for the endpoint (peer) metadata is added by Fishjam. Therefore, if we add such type definitions they should be in the `FishjamClient`, not `WebRTCEndpoint`.

Why? So far the `WebRTCEndpoint` has been used as an SDK for Membrane RTC Engine. If we add such changes, this effectively means that we no longer support SDK's for the RTC Engine.

## Motivation and Context

## Types of changes

      not work as expected)

### DIFF
--- a/packages/ts-client/src/webrtc/types.ts
+++ b/packages/ts-client/src/webrtc/types.ts
@@ -321,14 +321,8 @@ export interface Endpoint<EndpointMetadata, TrackMetadata> {
   /**
    * Any information that was provided in {@link WebRTCEndpoint.connect}.
    */
-  metadata: {
-    peer?: EndpointMetadata;
-    server: any;
-  };
-  rawMetadata: {
-    peer?: any;
-    server: any;
-  };
+  metadata: any;
+  rawMetadata: any;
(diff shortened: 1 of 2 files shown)


packages/ts-client/src/webrtc/types.ts
```diff
@@ -321,14 +321,8 @@ export interface Endpoint<EndpointMetadata, TrackMetadata> {
...
-  metadata: {
-    peer?: EndpointMetadata;
-    server: any;
-  };
-  rawMetadata: {
-    peer?: any;
-    server: any;
-  };
+  metadata: any;
+  rawMetadata: any;
```
czerwiukk, line 325: **note:** those type could be casted/validated in fishjam client

czerwiukk:
this apparently safe change broke the tests :C

kamil-stasiak (commented):
Based on this [PR](https://github.com/fishjam-cloud/web-client-sdk/pull/109/files) I found a few more places that mentions `peer` and `server`. Do we need to revert them?

image
image

PR #165, opened 2024-10-31
FCE 756 enable unit test and remove validators

kamil-stasiak:
## Description

- Generics were removed from `webRTCEndpoint.ts`.
- Metadata parsers were removed from `webRTCEndpoint.ts`.
- The `rawMetadata` and `metadataParsingError` fields were removed.
- Some unit tests were fixed.
- Unit and E2E tests that tested metadata parsers were deleted.
- Fishjam in E2E tests was updated to version 0.9.0.
- Unit tests were enabled to run on every PR.

## Motivation and Context
**Why were metadata parsers removed from `webRTCEndpoint.ts`?**

Because Fishjam changed the peer metadata structure. Fishjam introduced a separation between peer and server metadata. Previously, if a peer set its metadata to `{ a: "sth" }`, Fishjam would broadcast this change to other peers as the same object (`{ a: "sth" }`). Now, if the user sets the same object, Fishjam broadcasts it as `{ peer: { a: "sth" }, server: { b: "other server metadata" } }`. I didn't want to introduce more complicated parsers, so I deleted them.

## Types of changes


roznawsk (approved):
Let's get this merged!

packages/react-client/src/hooks/useVAD.ts
```diff
@@ -1,9 +1,13 @@
 import { useEffect, useMemo, useState } from "react";
-import type { PeerId, PeerMetadata, TrackId, TrackMetadata } from "../types/internal";
+import type { PeerId, TrackId } from "../types/internal";
 import { useFishjamContext } from "./useFishjamContext";
 import type { TrackContext, VadStatus } from "@fishjam-cloud/ts-client";
 import { useFishjamClientState } from "./useFishjamClientState";
 
+function isMicrophone(metadata: unknown) {
```
czerwiukk, line 7: **note:** this is not needed,  `metadata?.type === "microphone"` is all we need to make sure

mironiasty (approved):
LGTM

roznawsk (approved):

PR #172, opened 2024-11-08
Move WebRTC Endpoint to another package

czerwiukk:
## Description

WebRTC endpoint can be used without FishjamClient to work with Membrane RTC Engine alone.

## Types of changes

      not work as expected)


packages/webrtc-client/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@fishjam-cloud/webrtc-client",
+  "version": "0.9.0",
```
roznawsk, line 3: Maybe we should start versioning from `0.1.0`

packages/webrtc-client/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "description": "Typescript client library for Fishjam Cloud",
```
roznawsk, line 4:
```suggestion
  "description": "Typescript client library for ExWebRTC/WebRTC endpoint in Membrane RTC Engine",
```

roznawsk (approved):
LGTM

packages/webrtc-client/package.json
czerwiukk: why should we if the whole monorepo shares the same version?

PR #177, opened 2024-11-12
FCE-856/allow user to set metadata type on `usePeers`

czerwiukk:
## Description

Tidy up the metadata generic types.
Remove the `TrackMetadata` generic because we use a certain schema for track metadata.
Allow users to set metadata types using the `usePeers`  hook. 
When the metadata generics are not set, the type for both peer and server metadata is `Record<string, unknown>`.

## Types of changes

      not work as expected)

### DIFF
--- a/examples/react-client/fishjam-chat/src/components/RoomView.tsx
+++ b/examples/react-client/fishjam-chat/src/components/RoomView.tsx
@@ -5,7 +5,7 @@ import { CallToolbar } from "./CallToolbar";
 import { nonNullablePredicate } from "@/lib/utils";
 
 export const RoomView = () => {
-  const { localPeer, remotePeers } = usePeers();
+  const { localPeer, remotePeers } = usePeers<{ displayName: string }>();
 
   const trackAmount = [localPeer, ...remotePeers]
     .filter(nonNullablePredicate)
@@ -76,7 +76,7 @@ export const RoomView = () => {
                   )}
                 </Fragment>
               );
-            },
+            }
           )}
         </div>
       </section>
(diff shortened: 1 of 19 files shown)


packages/ts-client/src/types.ts
```diff
@@ -11,34 +11,41 @@ import type {
...
+  paused: boolean;
+  // track label used in recordings
+  displayName?: string;
 };
 
-type TrackContextEvents<Metadata> = {
-  encodingChanged: (context: FishjamTrackContext<Metadata>) => void;
-  voiceActivityChanged: (context: FishjamTrackContext<Metadata>) => void;
+export type Metadata<P, S> = {
+  peer: P;
```
roznawsk, line 22: Isn't peer metadata also optional?

roznawsk (approved):

packages/ts-client/src/types.ts
czerwiukk: Thanks for noticing, actually neither of those should be.
The default type shall be `Record<string, unknown> | undefined`

examples/react-client/fishjam-chat/src/components/RoomView.tsx
```diff
@@ -5,7 +5,7 @@ import { CallToolbar } from "./CallToolbar";
 import { nonNullablePredicate } from "@/lib/utils";
 
 export const RoomView = () => {
-  const { localPeer, remotePeers } = usePeers();
+  const { localPeer, remotePeers } = usePeers<{ displayName: string }>();
```
MiloszFilimowski, line 8: **suggestion:** I think you already have `PeerMetadata` declared in
- `packages/react-client/src/types/internal.ts`

Peer metadata are also set locally so there must be at least one other place where the same type is used 🤔

MiloszFilimowski (approved):

examples/react-client/fishjam-chat/src/components/RoomView.tsx
czerwiukk: good catch because that type should be deleted

PR #178, opened 2024-11-15
Update local endpoint metadata on connected

roznawsk:
## Description

Describe your changes in detail.

## Motivation and Context

## Types of changes

      not work as expected)

### DIFF
--- a/packages/webrtc-client/src/webRTCEndpoint.ts
+++ b/packages/webrtc-client/src/webRTCEndpoint.ts
@@ -97,14 +97,17 @@ export class WebRTCEndpoint extends (EventEmitter as new () => TypedEmitter<Requ
     switch (deserializedMediaEvent.type) {
       case 'connected': {
         this.local.setLocalEndpointId(deserializedMediaEvent.data.id);
+        this.local.setEndpointMetadata(deserializedMediaEvent.data.metadata);
 
         const endpoints = deserializedMediaEvent.data.otherEndpoints as EndpointWithTrackContext[];
 
         // todo implement track mapping (+ validate metadata)
         // todo implement endpoint metadata mapping
-        endpoints.forEach((endpoint) => {
-          this.remote.addRemoteEndpoint(endpoint);
-        });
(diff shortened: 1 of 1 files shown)


packages/webrtc-client/src/webRTCEndpoint.ts
```diff
@@ -97,14 +97,17 @@ export class WebRTCEndpoint extends (EventEmitter as new () => TypedEmitter<Requ
...
 
         const endpoints = deserializedMediaEvent.data.otherEndpoints as EndpointWithTrackContext[];
 
         // todo implement track mapping (+ validate metadata)
         // todo implement endpoint metadata mapping
-        endpoints.forEach((endpoint) => {
-          this.remote.addRemoteEndpoint(endpoint);
-        });
+        endpoints
+          .filter((endpoint) => endpoint.id != this.local.getEndpoint().id)
```
czerwiukk, line 107: **question:** Why is `data.otherEndpoints` prop expected to include the current endpoint id, which is presumably set in the 99th line as `data.id`?

czerwiukk: this one is just a better practice, non-blocking
```suggestion
          .filter((endpoint) => endpoint.id !== this.local.getEndpoint().id)
```

roznawsk: In the `otherEndpoints` list, we now also include the current endpoint (so that we can update is metadata). Tha's why the `endpointId` is contained there.

roznawsk:
This will be added along with the protobufs

packages/webrtc-client/src/webRTCEndpoint.ts
czerwiukk: but why is it in `otherEndpoints`?
can't we expose a interface like
```
type ConnectedEvent = { endpointId: string, endpointMetadata:  unknown, otherEndpoints: Endpoint[] }
```?

PR #186, opened 2024-11-19
FCE-883 / Bundle webrtc package in ts-client

czerwiukk:
## Description

Configure TSUP to bundle webrtc client in the ts client.
Additionally add `protobufs` package and store the protobufs there, instead of keeping them in `ts-client`.

## Motivation and Context

This allows to release `ts-client` without publishing `webrtc-client`.

## Types of changes

      not work as expected)


packages/ts-client/src/index.ts
```diff
@@ -18,6 +18,8 @@ export type { AuthErrorReason } from './auth.js';
 
 export { isAuthError, AUTH_ERROR_REASONS } from './auth.js';
 
+export { sdkVersion } from './sdkVersion';
```
mironiasty, line 21: Why this line was added?

mironiasty (commented):
How did you test that package will still work after publishing (and no files are missing?)

czerwiukk:
> How did you test that package will still work after publishing (and no files are missing?)

@mironiasty I used `yarn pack` to create a tarball, installed it via `yarn add package.tgz` and ran the example.

packages/ts-client/src/index.ts
czerwiukk: I recreated the protos submodule. There is no tags and I forgot to pull a 0.9.0 compatible commit.
Will undo.

Karolk99 (approved):

PR #198, opened 2024-11-26
Unify formatting configurations

mironiasty:
## Description

It looks like vscode does not always use default prettier configuration if priettier config is missing.

This PR adds missing files. And re-format example app to be in sync with react-client configuration.

## Motivation and Context

Let's have defined formatting config

## Types of changes

      not work as expected)


examples/react-client/fishjam-chat/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "printWidth": 120,
```
czerwiukk, line 2: **question:** why is the printWidth different than in different packages?

czerwiukk (approved):

PR #212, opened 2024-12-03
FCE-960 device api boildown

czerwiukk:
## Description

Removes excess of properties and functions.

## Motivation and Context

We strive for a simple tool for video conferences.
We will add methods if they're needed, but first let's come up with minimal setup.

## Types of changes

      not work as expected)


packages/react-client/src/types/public.ts
```diff
@@ -29,19 +29,44 @@ export type TracksMiddleware = (
...
+  activeDevice: DeviceItem | null;
   currentMiddleware: TrackMiddleware;
   deviceError: DeviceError | null;
   isMuted: boolean;
-  isDeviceEnabled: boolean;
-} & Omit<TrackManager, "currentTrack">;
+};
+
+export type MicrophoneApi = {
+  toggleMicrophone: () => void;
```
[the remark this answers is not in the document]
czerwiukk: added

examples/react-client/use-camera-and-microphone/src/AdditionalControls.tsx
```diff
@@ -40,29 +34,52 @@ export const AdditionalControls = () => {
...
-              status={peerStatus}
-            />
+            <div className="flex flex-col gap-2">
+              <button
+                className="btn btn-success btn-sm"
+                onClick={() => {
+                  camera.toggleCamera();
+                }}
+              >
+                Toggle camera (it's now {camera.isCameraOn ? "on" : "off"})
```
MiloszFilimowski, line 44: **question:** Is this correct?

czerwiukk: i'd say yes, this is to let the user know what will happen upon a click (earlier there was two buttons, for on and off)

MiloszFilimowski (approved):

PR #218, opened 2024-12-06
FCE-972 / lint and format once

czerwiukk:
## Description

Instead of running eslint and prettier for each package separately, just run once at the root level.

## Types of changes

      not work as expected)


mironiasty (approved):
LGTM 👍

PR #226, opened 2024-12-10
fix re-exporting from ts-client

czerwiukk:
## Description

Re-exports everything from `ts-client` as is.

## Motivation and Context

Variant enum has been accidentally exported just as a type, which makes it impossible to use it as a value.

## Types of changes

      not work as expected)


mironiasty (approved):
LGTM

PR #244, opened 2025-01-07
Fce 813/handle events while reconnecting

czerwiukk:
## Description

Implements a simple event queue to hold events while reconnecting.
Makes the reconnection enabled by default.

## Motivation and Context

The change mitigates the `websocket is still in CONNECTING state` error during the reconnection.

## Types of changes

      not work as expected)


packages/ts-client/src/FishjamClient.ts
```diff
@@ -87,8 +89,15 @@ export class FishjamClient<PeerMetadata = GenericMetadata, ServerMetadata = Gene
     this.reconnectManager = new ReconnectManager<PeerMetadata, ServerMetadata>(
       this,
       (peerMetadata) => this.initConnection(peerMetadata),
-      config?.reconnect,
+      config?.reconnect ?? true,
     );
+
+    this.eventQueue = new EventQueue({
```
czerwiukk, line 95: **thought:** this should be rather called peerMessageQueue

packages/ts-client/src/FishjamClient.ts
```diff
@@ -254,10 +260,20 @@ export class FishjamClient<PeerMetadata = GenericMetadata, ServerMetadata = Gene
...
   private setupCallbacks() {
     this.webrtc?.on('sendMediaEvent', (mediaEvent) => {
       const peerMediaEvent = PeerMediaEvent.decode(mediaEvent);
-      this.websocket?.send(PeerMessage.encode({ peerMediaEvent }).finish());
+      const serializedMessage = PeerMessage.encode({ peerMediaEvent }).finish();
+
+      if (this.isConnectingRelatedEvent(peerMediaEvent)) {
+        this.websocket?.send(serializedMessage);
+      } else {
+        this.peerMessageQueue.enqueueMessage(serializedMessage);
```
MiloszFilimowski, line 275: **question:** Just out of curiosity, what is the logic here? You enqueue every other event other those in `isConnectingRelatedEvent` and they only get sent when `reconnected` event is fired. What if it's never fired? 🤔

packages/ts-client/src/tests/messageQueue.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    // no trigger to send out yet
+    expect(sendMessage).not.toHaveBeenCalled();
+
+    const anotherMessage = new Uint8Array([7, 8, 9]);
+    queue.enqueueMessage(anotherMessage);
+
+    // all messages dispatched
+    expect(sendMessage).toHaveBeenCalledTimes(3);
+  });
+});
```
MiloszFilimowski, line 79: **praise:** Nice ✨

MiloszFilimowski (approved):

packages/ts-client/src/FishjamClient.ts
czerwiukk: The events related to establishing a connection are not enqueued because they are required for the reconnection to finish.
They would wait for the reconnection to finish and the `reconnected` event would never fire.

czerwiukk: If the reconnection fails due to network issues, the situation is screwed anyway.

packages/ts-client/src/messageQueue.ts
```diff
@@ -0,0 +1,34 @@
...
+  checkIsReconnecting: () => boolean;
+  sendMessage: (message: Uint8Array) => void;
+};
+
+export class MessageQueue {
+  private queuedMessages: Uint8Array[] = [];
+  private checkIsReconnecting: MessageQueueDeps['checkIsReconnecting'];
+  private sendMessage: MessageQueueDeps['sendMessage'];
+
+  constructor(deps: MessageQueueDeps) {
```
mironiasty, line 11: nit: I have different association with word `deps`, than function params (so I would rather call it `params`?) Or maybe just:
```suggestion
  constructor({checkIsReconnecting, sendMessage} : MessageQueueParams) {
```

mironiasty (approved):

packages/ts-client/src/messageQueue.ts
czerwiukk: done, what's the association though? i meant dependencies

PR #247, opened 2025-01-15
[FCE-1038] Update types exports (and typedoc) to make `yarn docs` happy

mironiasty:
## Description

Export all types that are visible to end user, but were not yet exported
Fix typedoc syntax marked as incorrect
Update typedoc config to avoid part of validation for libraries that are not actively supported

## Motivation and Context

I want to update our documentation, but as part of this process I want to remove warnings (and also add missing types into docs)

## Types of changes

      not work as expected)

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -45,3 +45,6 @@ jobs:
 
       - name: Run typecheck 🚓
         run: yarn tsc
+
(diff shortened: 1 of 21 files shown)


czerwiukk (approved):
lgtm and i've even checked - it truly does eliminate the warnings

PR #250, opened 2025-01-20
[FCE-1057] Add branded PeerId and TrackId

mironiasty:
## Description

Add branded types to most common IDs used in React Client. This way, these IDs can't be used as strings.

## Motivation and Context

Make client code a little less buggy

## Types of changes

      not work as expected)


czerwiukk (approved):
Looks good. I presume hooks that takes `PeerId` as an argument (like `useVAD`) affected immediately and now accept branded strings only.

mironiasty:
Yes, this change affects all usages of PeerId type from react client (ts client is not affected, so strings are still allowed when someone will use this library directly)

PR #251, opened 2025-01-20
FCE-1059/fix reconnection manager endpoint metadata retrieval

czerwiukk:
## Description

Fixes low level peer metadata retrieval from RTC Endpoint.

## Motivation and Context

The tsclient's `ReconnectionManager` was still unaware of the metadata being divided between peer and server.
The type of the metadata at the `LocalTrack` level is `unknown` and `ReconnectionManager` was wrongfully casting it as `PeerMetadata`.
This led to issues with providing correct metadata while reconnecting.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/ts-client/src/reconnection.ts
+++ b/packages/ts-client/src/reconnection.ts
@@ -46,7 +46,7 @@ export class ReconnectManager<PeerMetadata, ServerMetadata> {
 
   private readonly connect: (metadata: PeerMetadata) => void;
   private readonly client: FishjamClient<PeerMetadata, ServerMetadata>;
-  private initialMetadata: PeerMetadata | undefined | null = undefined;
+  private initialPeerMetadata: PeerMetadata | undefined | null = undefined;
 
   private reconnectAttempt: number = 0;
   private reconnectTimeoutId: NodeJS.Timeout | null = null;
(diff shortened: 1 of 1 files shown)


packages/ts-client/src/reconnection.ts
```diff
@@ -96,15 +96,18 @@ export class ReconnectManager<PeerMetadata, ServerMetadata> {
...
+    this.initialPeerMetadata = initialPeerMetadata;
     this.reconnectAttempt = 0;
     if (this.reconnectTimeoutId) clearTimeout(this.reconnectTimeoutId);
     this.reconnectTimeoutId = null;
   }
 
   private getLastPeerMetadata(): PeerMetadata | undefined {
-    return this.lastLocalEndpoint?.metadata as PeerMetadata;
+    const endpointMetadata = this.lastLocalEndpoint?.metadata as
+      | { peer: PeerMetadata; server: ServerMetadata }
```
mironiasty, line 108:
```suggestion
      | Metadata<PeerMetadata, ServerMetadata>
```

https://github.com/fishjam-cloud/web-client-sdk/blob/af10e28c81475a654848b55658f60509693cf641/packages/ts-client/src/types.ts#L30

mironiasty (approved):
probably better type can be used

packages/ts-client/src/reconnection.ts
czerwiukk: yup! that's how it should be

PR #259, opened 2025-01-22
[FCE-1082] - Fix generated output of TrackContextImpl

MiloszFilimowski:
## Description

- Refactored TrackContextImpl to avoid runtime errors caused by useDefineForClassFields.

## Motivation and Context

Our tsconfig either enables `useDefineForClassFields` or sets `target: es2022` (enabling it by default). This can cause issues when subclassing, as this may be used before `super()` is called. More details [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier). 
This caused runtime errors with `WebRTCEndpoint`, especially in Playwright tests for `membrane-rtc-engine`.

We have two options:

1. Disable `useDefineForClassFields`, which is risky and could impact the codebase.
2. Refactor `TrackContextImpl` using declare to modify JS output and move initialization to the constructor.

I chose option 2 as it is less invasive and resolves the issue safely.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/webrtc-client/src/internal.ts
+++ b/packages/webrtc-client/src/internal.ts
@@ -20,23 +20,20 @@ export class TrackContextImpl
   extends (EventEmitter as new () => TypedEmitter<Required<TrackContextEvents>>)
   implements TrackContext
 {
-  endpoint: Endpoint;
-  trackId: string;
-  track: MediaStreamTrack | null = null;
-  trackKind: TrackKind | null = null;
-  stream: MediaStream | null = null;
-  metadata?: unknown;
-  metadataParsingError?: any;
-  simulcastConfig?: MediaEvent_Track_SimulcastConfig;
-  maxBandwidth: TrackBandwidthLimit = 0;
-  encoding?: Variant;
-  encodingReason?: EncodingReason;
-  vadStatus: VadStatus = 'silence';
-  negotiationStatus: TrackNegotiationStatus = 'awaiting';
-
-  // Indicates that metadata were changed when in "offered" negotiationStatus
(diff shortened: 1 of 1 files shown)


packages/webrtc-client/src/internal.ts
```diff
@@ -20,23 +20,22 @@ export class TrackContextImpl
...
+  declare simulcastConfig?: MediaEvent_Track_SimulcastConfig;
+  declare maxBandwidth: TrackBandwidthLimit;
+  declare encoding?: Variant;
+  declare encodingReason?: EncodingReason;
+  declare vadStatus: VadStatus;
+  declare negotiationStatus: TrackNegotiationStatus;
   // Indicates that metadata were changed when in "offered" negotiationStatus
   // and `updateTrackMetadata` Media Event should be sent after the transition to "done"
-  pendingMetadataUpdate: boolean = false;
+  declare pendingMetadataUpdate: boolean;
```
roznawsk, line 38: **question:** What is the meaning of `declare` keyword?

roznawsk (approved):

packages/webrtc-client/src/internal.ts
MiloszFilimowski: >`declare` is used to tell the compiler "this thing" exists already, and therefore can be referenced by other code, also there is no need to compile this statement into any JavaScript".

So in this case, because this is a subclass it's only used to change property type and won't emit any JS. This is what this bug was about.

czerwiukk (approved):
feels a bit hacky, but it's fine if fixes the issue

PR #267, opened 2025-01-30
[FCE-1112]: Better information on connected peer limit reached

MiloszFilimowski:
## Description

- Always log a warning with the reason when the socket is closed
- Emit join error when a known error occurres.

## Motivation and Context

- Socket errors were silently ignored and resulted in an unknown state. 

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/index.ts
+++ b/packages/react-client/src/index.ts
@@ -24,6 +24,7 @@ export type {
 } from "./types/public";
 export type {
   AuthErrorReason,
+  JoinErrorReason,
   Metadata,
   ReconnectConfig,
   ReconnectionStatus,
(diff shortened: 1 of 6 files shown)


mironiasty (approved):
I'm not a big fan 🪭 of `joinError` name 🤔 

But I don't have suggestion for perfect name, also it would require bigger refactor, so let's merge it 🤷

czerwiukk (approved):

czerwiukk:
> I'm not a big fan 🪭 of `joinError` name 🤔
> 
> But I don't have suggestion for perfect name, also it would require bigger refactor, so let's merge it 🤷

I can suggest `JoiningRoomError` from the top of my head.

I think it'd be best that those errors were an enum generated from the protobufs, but that'd be a breaking change and too much work for the current moment.
The PR introduce improvements and let's merge it fast.

PR #292, opened 2025-03-03
Fce 1279/patch race conditions and poor cleanup

czerwiukk:
## Description

This PR mitigates the issue of starting multiple streams and being unable to stop them later.
The camera is now working correctly in `React.StrictMode`.

Please aware that this is rather **a hack** and a rewrite of the Device Manager is already planned.

## Motivation and Context

Due to StrictMode nature, the DeviceManager's `start` function was called twice, leading to 2 streams being retrieved from the browser, while only one was manageable.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/FishjamProvider.tsx
+++ b/packages/react-client/src/FishjamProvider.tsx
@@ -53,7 +53,8 @@ export interface FishjamProviderProps extends PropsWithChildren {
 export function FishjamProvider(props: FishjamProviderProps) {
   const fishjamClientRef = useRef(new FishjamClient({ reconnect: props.reconnect }));
 
-  const hasDevicesBeenInitializedRef = useRef(false);
+  const devicesInitializationRef = useRef<Promise<void> | null>(null);
+
   const storage = props.persistLastDevice;
 
   const videoDeviceManagerRef = useRef(
@@ -84,6 +85,7 @@ export function FishjamProvider(props: FishjamProviderProps) {
     getCurrentPeerStatus,
     bandwidthLimits: mergedBandwidthLimits,
     streamConfig: props.videoConfig,
+    devicesInitializationRef,
   });
 
   const audioTrackManager = useTrackManager({
@@ -92,6 +94,7 @@ export function FishjamProvider(props: FishjamProviderProps) {
     getCurrentPeerStatus,
     bandwidthLimits: mergedBandwidthLimits,
     streamConfig: props.audioConfig,
+    devicesInitializationRef,
   });
 
   const screenShareManager = useScreenShareManager({ fishjamClient: fishjamClientRef.current, getCurrentPeerStatus });
@@ -106,7 +109,7 @@ export function FishjamProvider(props: FishjamProviderProps) {
     audioTrackManager,
     videoDeviceManagerRef,
     audioDeviceManagerRef,
-    hasDevicesBeenInitializedRef,
+    devicesInitializationRef,
     clientState,
     bandwidthLimits: mergedBandwidthLimits,
   };
--- a/packages/react-client/src/devices/DeviceManager.ts
+++ b/packages/react-client/src/devices/DeviceManager.ts
@@ -242,6 +242,7 @@ export class DeviceManager
     this.middlewareManager.clearMiddleware();
     this.processedMediaTrack?.stop();
     this.rawMedia?.track?.stop();
+    this.media?.track?.stop();
 
     this.updateMedia(null);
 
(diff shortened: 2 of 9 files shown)


packages/react-client/src/FishjamProvider.tsx
```diff
@@ -53,7 +53,8 @@ export interface FishjamProviderProps extends PropsWithChildren {
 export function FishjamProvider(props: FishjamProviderProps) {
   const fishjamClientRef = useRef(new FishjamClient({ reconnect: props.reconnect }));
 
-  const hasDevicesBeenInitializedRef = useRef(false);
+  const devicesInitializationRef = useRef<Promise<void> | null>(null);
```
czerwiukk, line 56: **note:** this is indeed a hack, but it allowed to mitigate the race condition between fetching streams in `useInitializeDevices` and `start` method in the `DeviceManager`

packages/react-client/src/devices/DeviceManager.ts
```diff
@@ -178,22 +178,8 @@ export class DeviceManager
...
-      // Safari always deactivates the track and emits the `ended` event.
-      // Its handling is asynchronous and can be executed even before returning a value from the re-execution of `getUserMedia`.
-      // In such a case, the tracks are already deactivated at this point (logic in `onTrackEnded` method).
-      // The track is null, so the stop method will not execute.
-      //
-      // However, if Safari has not yet handled this event, the tracks are manually stopped at this point.
-      // Manually stopping tracks on its own does not generate the `ended` event.
-      // The ended event in Safari has already been emitted and will be handled in the future.
-      // Therefore, in the `onTrackEnded` method, events for already stopped tracks are filtered out to prevent the state from being damaged.
-      if (shouldReplaceDevice) {
```
czerwiukk, line 193: **note:** stopping stale tracks only while changing a device led to being unable to disable them at all later, so the camera led was always on

packages/react-client/src/hooks/devices/useMicrophone.ts
```diff
@@ -14,6 +14,10 @@ export function useMicrophone() {
     toggleMicrophone: audioTrackManager.toggleDevice,
     /** Mutes/unmutes the microphone */
     toggleMicrophoneMute: audioTrackManager.toggleMute,
+    /** Toggles current microphone on */
+    enableMicrophone: audioTrackManager.enableDevice,
+    /** Toggles current microphone off */
+    disableMicrophone: audioTrackManager.disableDevice,
```
mironiasty, line 20: Same question applies here

packages/react-client/src/utils/deferred.ts
```diff
@@ -0,0 +1,21 @@
+export class Deferred<T, E = unknown> {
+  public promise: Promise<T>;
```
mironiasty, line 2: Could this be marked as readonly?

packages/react-client/src/hooks/devices/useInitializeDevices.ts
```diff
@@ -17,13 +18,16 @@ export type InitializeDevicesErrors = { audio: DeviceError | null; video: Device
...
   const initializeDevices: (params?: UseInitializeDevicesParams) => Promise<null | InitializeDevicesErrors> =
     useCallback(
       async ({ enableVideo = true, enableAudio = true }: UseInitializeDevicesParams = {}) => {
-        if (hasDevicesBeenInitializedRef.current) return null;
-        hasDevicesBeenInitializedRef.current = true;
+        if (devicesInitializationRef.current) return null;
+
+        const deferred = new Deferred<void>();
+
+        devicesInitializationRef.current = deferred.promise;
```
mironiasty, line 30: Should logic in useCallback be wrapped in try/catch. So `deferred.promise` could be rejected if something will fall?

packages/react-client/src/hooks/internal/useTrackManager.ts
```diff
@@ -137,15 +139,51 @@ export const useTrackManager = ({
     }
   }, [tsClient, getCurrentPeerStatus, resumeStreaming, startStreaming]);
 
+  const enableDevice = useCallback(async () => {
+    if (devicesInitializationRef.current) {
```
mironiasty, line 143: Is this condition really required?

packages/react-client/src/hooks/devices/useCamera.ts
```diff
@@ -14,6 +14,14 @@ export function useCamera() {
      * Toggles current camera on/off
      */
     toggleCamera: videoTrackManager.toggleDevice,
+    /**
+     * Toggles current camera on
+     */
+    enableCamera: videoTrackManager.enableDevice,
```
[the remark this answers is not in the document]
MiloszFilimowski: I agree. Do we need this now? Or maybe it can wait for the proper fix?

PR #304, opened 2025-03-11
FCE-1278 device manager overhaul

czerwiukk:
## Description

Gets rid of the DeviceManager classes and puts everything in the React lifecycle.

## Motivation and Context

Glueing the old DeviceManager class with the `useInitializeDevices` was resulting in race conditions regarding retrieving streams from the browser.
Additionally, we've finally unified the whole `react-client` to one code style.

## Types of changes

      not work as expected)


packages/react-client/src/hooks/internal/device/useDevice.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+function stopStream(stream: MediaStream, type: "audio" | "video") {
+  getCertainTypeTracks(stream, type).forEach((track) => {
+    track.enabled = false;
+    track.stop();
+  });
+}
+
+function getReplaceStreamAction(
```
MiloszFilimowski, line 50: **question:** What does it do? `getReplaceStreamAction` is very enigmatic

packages/react-client/src/hooks/internal/device/useDevice.ts
```diff
@@ -0,0 +1,186 @@
...
+  if (deviceId) {
+    constraints.deviceId = { exact: deviceId };
+  }
+  const stream = await navigator.mediaDevices.getUserMedia({
+    [type]: constraints,
+  });
+  return stream;
+}
+
+export const useDevice = ({
```
MiloszFilimowski, line 90: **suggestion:** Do you think this function could be simplified more? There is a lof of going on and it's difficult to understand at a high level what it does. 

Do you think it would be possible to create more abstractions over the existing code? And maybe even split this file into more files?

packages/react-client/src/hooks/internal/device/useDevices.ts
```diff
@@ -0,0 +1,140 @@
...
+
+interface UseDevicesProps {
+  videoConstraints?: MediaTrackConstraints | boolean;
+  audioConstraints?: MediaTrackConstraints | boolean;
+  persistHandlers?: PersistLastDeviceHandlers;
+}
+
+type InitializeDevicesResult = { stream: MediaStream | null; errors: AudioVideo<DeviceError | null> | null };
+
+export const useDevices = ({ videoConstraints, audioConstraints, persistHandlers }: UseDevicesProps) => {
```
MiloszFilimowski, line 17: **suggestion:** I'd also try to simplify this function if that's possible. `initializeDevices` sounds like a great candidate for extraction.

packages/react-client/src/hooks/internal/useFishjamClient_DO_NOT_USE.ts
```diff
@@ -1,14 +1,14 @@
...
 
-import { useFishjamContext } from "./useFishjamContext";
+import { FishjamClientContext } from "../../contexts/fishjamClient";
 
 /**
  *
  * @category Connection
  * @deprecated
  */
 export function useFishjamClient_DO_NOT_USE() {
```
MiloszFilimowski, line 10: **question:** Why can't we just remove this altogether? I thought it was only needed for the old video room?

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: it's used like that `setStream(getReplaceStreamAction(stream, deviceType))`
it's a function of `SetStateAction` type,  the one that's passed to a state setter and takes the old state as an argument.
i use it to always clean up the old stream while setting a new one

packages/react-client/src/hooks/devices/useCamera.ts
```diff
@@ -1,13 +1,22 @@
...
 
 /**
  * This hook can toggle camera on/off, change camera, provides current camera and other.
  * @category Devices
  */
 export function useCamera() {
-  const { videoTrackManager, videoDeviceManagerRef } = useFishjamContext();
-  const deviceApi = useDeviceApi({ deviceManager: videoDeviceManagerRef.current });
+  const cameraCtx = useContext(CameraContext);
+  if (!cameraCtx) throw Error("useCamera must be used within CameraProvider");
```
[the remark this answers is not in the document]
czerwiukk: i think in this case linter won't scream either way - the condition is placed after all of the hooks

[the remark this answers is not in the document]
czerwiukk: we can put the rule, either way, it's a story for another PR

packages/react-client/src/hooks/internal/device/useDevices.ts
czerwiukk: `initializeDevices` is the only relevant function that this hook have - it uses the majority of the state that's also used by the already extracted `useDevice`. Don't you think it'd be an overkill?

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: I think useDevice is not the best name. I will think about it.This hook contains all of the needed functions to interact with a device of a certain type - mic or camera. Middleware hook is already extracted.

examples/react-client/fishjam-chat/src/components/CallToolbar.tsx
```diff
@@ -29,7 +29,20 @@ export const CallToolbar = () => {
 
   const { toggleCamera, isCameraOn } = useCamera();
 
-  const { toggleMicrophoneMute, isMicrophoneMuted } = useMicrophone();
+  const {
+    toggleMicrophoneMute,
+    isMicrophoneMuted,
+    toggleMicrophone,
+    isMicrophoneOn,
+  } = useMicrophone();
```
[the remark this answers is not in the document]
czerwiukk: [FCE-1345](https://linear.app/swmansion/issue/FCE-1345/consider-simplifying-the-microphone-hardsoft-mute-with-mobile-sdk)

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: changed it to `useDeviceManager`

packages/react-client/src/hooks/internal/useFishjamClient_DO_NOT_USE.ts
czerwiukk: [tickecik](https://linear.app/swmansion/issue/FCE-1349/remove-usefishjamclient-do-not-use)

packages/react-client/src/hooks/internal/device/useDevice.ts
```diff
@@ -0,0 +1,172 @@
...
+  constraints,
+  saveUsedDevice,
+  deviceError,
+  setError,
+}: DeviceManagerProps): DeviceManager => {
+  const rawTrack = useMemo(() => mediaStream && getTrackFromStream(mediaStream, deviceType), [mediaStream, deviceType]);
+
+  const { processedTrack, applyMiddleware, currentMiddleware } = useTrackMiddleware(rawTrack);
+
+  const currentTrack = useMemo(() => processedTrack ?? rawTrack, [rawTrack, processedTrack]);
```
mironiasty, line 90: I don't think `useMemo` is required here

packages/react-client/src/hooks/internal/device/useDevice.ts
```diff
@@ -0,0 +1,172 @@
...
+      );
+    if (!currentDevice) return null;
+    return { label: currentDevice.label, deviceId: currentDevice.deviceId };
+  }, [mediaStream, deviceList, deviceType]);
+
+  const [deviceEnabled, setDeviceEnabled] = useState(true);
+
+  const startDevice = useCallback(
+    async (deviceId?: string) => {
+      let stream = await getInitialStream();
```
mironiasty, line 111: Does `stream` variable needs to be overwritten?
```suggestion
      const initialStream = await getInitialStream();
```

packages/react-client/src/hooks/internal/device/useDevice.ts
```diff
@@ -0,0 +1,172 @@
...
+export const useDeviceManager = ({
+  mediaStream,
+  getInitialStream,
+  deviceType,
+  allDevicesList,
+  setStream,
+  constraints,
+  saveUsedDevice,
+  deviceError,
+  setError,
```
mironiasty, line 84: Based on usage, I would consider one of two changes here:
- rename `setStream` and `setError` to `setMediaStream` and `setDeviceError`.

Or hide `mediaStream` and `deviceError` as internal `useDeviceManager` state. And return setters to be used in `useDevices`. This would also imply that `useHandleTrackEnd` would be called inside `useDeviceManager`

packages/react-client/src/hooks/internal/device/useDevice.ts
mironiasty, line 1: Should this file be renamed to `useDeviceManager`?

packages/react-client/src/hooks/internal/device/useDevices.ts
mironiasty, line 1: `useMediaDevices.ts`?

packages/react-client/src/hooks/internal/device/useDevices.ts
```diff
@@ -0,0 +1,144 @@
...
+      lastUsedMicRef.current = deviceInfo;
+      persistHandlers?.saveLastDevice(deviceInfo, "audio");
+    },
+    [persistHandlers],
+  );
+
+  const initializeDevices = useCallback(
+    async (settings?: { enableVideo?: boolean; enableAudio?: boolean }) => {
+      if (deviceList.length) {
+        return null;
```
mironiasty, line 51: Should we have different return value for cases `there are no devices` and `all is ok`?

packages/react-client/src/hooks/internal/device/useDevices.ts
```diff
@@ -0,0 +1,144 @@
...
+      initializationRef.current = null;
+      if (result.errors.video || result.errors.audio) return result.errors;
+      return null;
+    },
+    [deviceList, videoConstraints, audioConstraints, saveUsedCamera, saveUsedMic],
+  );
+
+  const getInitialStream = useCallback(async () => {
+    const result = await initializationRef.current;
+    return result?.stream ?? null;
```
mironiasty, line 109: Is this expected that this method return non-null value only when `initializeDevices` is still running?

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: that's correct

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: yes, as well as use media devices

packages/react-client/src/hooks/internal/device/useDevice.ts
czerwiukk: I'd like to hide `mediaStream` as well as `deviceError` inside a device manager, but I need this state higher up to be able to ask for both camera and microphone from the browser during initialization and then keep the streams I've obtained.

packages/react-client/src/hooks/internal/device/useDevices.ts
czerwiukk: I introduced a result object with a `status` property that describes the outcome.

packages/react-client/src/types/public.ts
```diff
@@ -1,6 +1,12 @@
 import type { SimulcastConfig, TrackMetadata, Variant } from "@fishjam-cloud/ts-client";
 
-import type { DeviceManagerStatus } from "./internal";
+export type InitializeDevicesStatus = "initialized" | "failed" | "initialized_with_errors" | "already_initialized";
+
+export type InitializeDevicesResult = {
+  status: InitializeDevicesStatus;
+  stream: MediaStream | null;
```
czerwiukk, line 7: **thought:** @mironiasty do you think it's ok to also return the retrieved stream as a result of the `initializeDevices` function? I added it now because it was less effort to leave it like that instead of removing the stream from the result object before returning it to the user (I need it internally).

mironiasty: I guess it should be fine 🤔

packages/react-client/src/hooks/internal/devices/useDeviceManager.ts
```diff
@@ -0,0 +1,172 @@
...
+export const useDeviceManager = ({
+  mediaStream,
+  getInitialStream,
+  deviceType,
+  allDevicesList,
+  setStream,
+  constraints,
+  saveUsedDevice,
+  deviceError,
+  setError,
```
mironiasty, line 84: If you want to keep all items on that list, I would consider renaming as mentioned in first point here:
https://github.com/fishjam-cloud/web-client-sdk/pull/304#discussion_r1995785065

mironiasty (approved):
One small comment left

MiloszFilimowski (approved):
💯

PR #305, opened 2025-03-14
Remove `useFishjamClient_DO_NOT_USE`

czerwiukk:
## Description

Removes deprecated hook and also an example that bases on it.

## Motivation and Context

It's no longer needed.

## Types of changes

      not work as expected)


examples/react-client/minimal-react/src/components/App.tsx
```diff
@@ -2,8 +2,8 @@ import {
   useConnection,
   usePeers,
   useScreenShare,
+  useStatistics,
```
MiloszFilimowski, line 5: **suggestion:** In mobile clinet we've hidden helpers like this under a nested package so it's available at `@fishjam-cloud/react-client/dev` Do you think it would make sense here as well?

MiloszFilimowski (approved):

examples/react-client/minimal-react/src/components/App.tsx
czerwiukk: yes! `/debug` makes sense

mironiasty (approved):

PR #313, opened 2025-03-18
FCE-1332 tidy up errors in track manager

czerwiukk:
## Description

Removes unnecessary logic from track manager.

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/hooks/internal/useTrackManager.ts
+++ b/packages/react-client/src/hooks/internal/useTrackManager.ts
@@ -37,7 +37,9 @@ export const useTrackManager = ({
 
   const selectDevice = useCallback(
     async (deviceId?: string) => {
-      const newTrack = await startDevice(deviceId);
+      const [newTrack, error] = await startDevice(deviceId);
+      if (error) return error;
+
       const currentTrackId = getCurrentTrackId();
       if (!currentTrackId) return;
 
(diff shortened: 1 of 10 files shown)


packages/react-client/src/hooks/internal/useTrackManager.ts
```diff
@@ -86,14 +79,10 @@ export const useTrackManager = ({
         const remoteTrackId = await tsClient.addTrack(track, trackMetadata, simulcastConfig, maxBandwidth);
 
         currentTrackIdRef.current = remoteTrackId;
-        setPaused(false);
-
-        return remoteTrackId;
       } catch (err) {
         if (err instanceof TrackTypeError) {
```
czerwiukk, line 83: **note:** this is fine, `tsClient.addTrack` only throws with `TrackTypeError`

packages/react-client/src/hooks/internal/useTrackManager.ts
```diff
@@ -150,7 +140,10 @@ export const useTrackManager = ({
       }
     } else {
       const newTrack = await startDevice();
-      if (!newTrack) throw Error("Device is unavailable");
+      if (!newTrack) {
+        console.warn(`Failed to start ${type} device.`);
+        return;
```
mironiasty, line 145: Should this info be somehow passed back to user as return value? I would expect that in production app warnings are removed from code, so there is now way to catch this issue.

czerwiukk: Now `toggleDevice` and `selectDevice` return a `Promise<undefined | DeviceError>` type.

mironiasty (approved):
👍

PR #325, opened 2025-04-03
Keep the devices disabled/muted state after changing device.

czerwiukk:
## Description

Preserves the initial state of device being turned/muted on/off after changing a device.

## Motivation and Context

Current solution of the devices being always turned on during changing them is prone to user sending out data he possibly wouldn't like to.

## Types of changes

      not work as expected)

### DIFF
--- a/examples/react-client/fishjam-chat/src/components/DeviceSettings.tsx
+++ b/examples/react-client/fishjam-chat/src/components/DeviceSettings.tsx
@@ -1,18 +1,36 @@
 import { useCamera, useMicrophone } from "@fishjam-cloud/react-client";
+import { Mic, MicOff, Video, VideoOff } from "lucide-react";
 
 import AudioVisualizer from "./AudioVisualizer";
 import { BlurToggleButton } from "./BlurToggle";
 import { DeviceSelect } from "./DeviceSelect";
+import { ToggleButton } from "./ToggleButton";
 import VideoPlayer from "./VideoPlayer";
 
 export const CameraSettings = () => {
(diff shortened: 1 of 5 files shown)


examples/react-client/fishjam-chat/src/components/DeviceSettings.tsx
```diff
@@ -1,18 +1,36 @@
...
+    cameraDevices,
+    selectCamera,
+    activeCamera,
+    toggleCamera,
+    isCameraOn,
+  } = useCamera();
 
   const hasValidDevices = cameraDevices.some((device) => device.deviceId);
 
+  const onCameraToggle = () => {
```
MiloszFilimowski, line 22: **question:** Is this needed? You can just pass `toggleCamera` to `onClick`. Same as with  `toggleMicrophone`.

packages/react-client/src/hooks/internal/devices/useDeviceManager.ts
```diff
@@ -79,6 +80,8 @@ export const useDeviceManager = ({
   deviceError,
   setDeviceError,
 }: DeviceManagerProps): DeviceManager => {
+  const selectedDeviceRef = useRef<string | undefined>(undefined);
```
MiloszFilimowski, line 83: **question:** Why undefined instead of null?

czerwiukk: Because `startDevice` takes `string | undefined` as an argument and I didn't want to handle another case.
While revisiting it now it think it might be worth it.

MiloszFilimowski (approved):

PR #326, opened 2025-04-04
FCE-1440 / unit testing setup

czerwiukk:
## Description

Installed `vitest` + some utils, added example test and did a small refactor.

## Motivation and Context

We had no unit test setup, this is the minimum and a base for further test writing.

## Types of changes

      not work as expected)


packages/react-client/src/hooks/internal/devices/useDeviceManager.ts
```diff
@@ -84,6 +85,8 @@ export const useDeviceManager = ({
 
   const rawTrack = useMemo(() => mediaStream && getTrackFromStream(mediaStream, deviceType), [mediaStream, deviceType]);
 
+  useHandleTrackEnd({ track: rawTrack, clearStream: () => setMediaStream(null) });
```
MiloszFilimowski, line 88: **question:** Won't this make the underlying `useEffect` be called on every rerender?

czerwiukk: The object is deconstructed inside the hook, so I thought the references passed to underlying are going to be exactly the same.
But then I saw this arrow function happily being recreated every render x-D

MiloszFilimowski (approved):

PR #334, opened 2025-04-09
FCE-1461 / Fix freshly fetched devices selection

czerwiukk:
## Description

Marking the devices of a new stream as selected was broken due to using `devicesList` state instead of the result of `enumerateDevices`.

## Types of changes

      not work as expected)


MiloszFilimowski (approved):

packages/react-client/src/hooks/internal/devices/useMediaDevices.ts
```diff
@@ -39,6 +39,12 @@ export const useMediaDevices = ({ videoConstraints, audioConstraints, persistHan
     [persistHandlers],
   );
 
+  const fetchDevices = useCallback(async () => {
+    const devices = await navigator.mediaDevices.enumerateDevices();
+    setDeviceList(devices);
```
MiloszFilimowski, line 44: **issue:** IMO this function should not setDeviceList, as it violates it's SRP, it's called `fetchDevices` so it should not manipulate the state. IMO I'd either call it `setDevices` or I'd move `setDeviceList` call to `intitialize` function.

czerwiukk: fair

PR #380, opened 2025-07-01
FCE-1638 connecting using Fishjam ID

czerwiukk:
## Description

This PR introduces:
- the `fishjamId` prop to the `FishjamProvider`
- the `useSandbox` hook

The `useSandbox` hook is a tool to use room manager without calling any urls by yourself.
Simply call `getSandboxPeerToken/ViewerToken` and receive the desired token.

All of the changes are additive, there is no breaking changes.
You can still pass a url to the `joinRoom` method of `useConnection` hook to override the url derived from the `fishjamId`.
The same goes for the `useSandbox` prop - it also takes an optional argument `roomManagerUrl`.

## Motivation and Context

This change allows users to set the Fishjam ID once, then connect to Fishjam and Room Manager without using any other ids nor urls.

## Types of changes

      not work as expected)

### DIFF
--- a/examples/react-client/livestreaming/src/components/ConnectForm.tsx
+++ b/examples/react-client/livestreaming/src/components/ConnectForm.tsx
@@ -1,5 +1,5 @@
-import { useForm } from "react-hook-form";
 import { useState } from "react";
+import { useForm } from "react-hook-form";
 
 import { useFishjamId } from "@/lib/fishjamContext";
 
@@ -23,12 +23,10 @@ export const ConnectForm = () => {
   });
 
   function onSubmit(values: FormValues) {
-    console.log("onSubmit", values);
     setIsLocked(true);
     setFishjamId(values.fishjamId);
   }
 
-  console.log("is locked", isLocked);
   return (
     <Card className="flex-grow mt-4 w-full max-w-lg">
       <CardContent className="space-y-2">
@@ -55,7 +53,6 @@ export const ConnectForm = () => {
                 type="button"
                 variant="secondary"
                 onClick={() => {
-                  console.log("X");
                   setIsLocked(false);
                 }}
               >


examples/react-client/fishjam-chat/src/main.tsx
```diff
@@ -10,7 +10,7 @@ import { Toaster } from "./components/ui/toaster";
 
 ReactDOM.createRoot(document.getElementById("root")!).render(
   <React.StrictMode>
-    <FishjamProvider>
+    <FishjamProvider fishjamId={"67eaef1141154713b4b2b58d7d54c16e"}>
```
MiloszFilimowski, line 13: **question:** Should this be harcoded?

MiloszFilimowski (approved):
Looks good. I'm not yet sure if we're going to implment this using a context on mobile but either way should be fine! GJ!

examples/react-client/fishjam-chat/src/main.tsx
AHGIJMKLKKZNPJKQR: **answer**: It shouldn't

packages/react-client/src/FishjamProvider.tsx
```diff
@@ -52,6 +53,11 @@ export interface FishjamProviderProps extends PropsWithChildren {
    * Configure whether to use audio simulcast and which layers to send if so.
    */
   audioConfig?: StreamConfig;
+  /*
+   * Your Fishjam app ID
```
AHGIJMKLKKZNPJKQR, line 57: **thought:** I think we should settle on a naming scheme for the tokens/ids before merging this.
Maybe it would be better to say flat out that this is either your sandbox id or prod id?
The user would then have 3 (or 4) things:
1. sandbox id
2. sandbox api token
3. prod api token
4. prod id (maybe not necessary)?

I'm getting the feeling that getting rid of the word "app" in general might be a good idea

packages/react-client/src/hooks/useConnection.ts
```diff
@@ -9,7 +10,7 @@ export interface JoinRoomConfig<PeerMetadata extends GenericMetadata = GenericMe
   /**
    * Fishjam URL
    */
-  url: string;
+  url?: string;
```
AHGIJMKLKKZNPJKQR, line 13: **suggestion:** I would change this field's description to something like "overrides the default URL derived from fishjamId".

packages/react-client/src/FishjamProvider.tsx
```diff
@@ -52,6 +53,11 @@ export interface FishjamProviderProps extends PropsWithChildren {
...
+   * This property will allow you to use `useSandbox` and `useConnection` hooks without passing any urls.
+   */
+  fishjamId?: string;
```
AHGIJMKLKKZNPJKQR, line 60: **suggestion:** What do you think about making this field required?

packages/react-client/src/hooks/useConnection.ts
```diff
@@ -39,8 +41,16 @@ export function useConnection() {
       url,
       peerToken,
       peerMetadata,
-    }: JoinRoomConfig<PeerMetadata>) => client.connect({ url, token: peerToken, peerMetadata: peerMetadata ?? {} }),
-    [client],
+    }: JoinRoomConfig<PeerMetadata>) => {
+      if (!url && !fishjamId) {
+        throw Error(
+          `You haven't passed the fishjamId to the FishjamProvider, nor have you passed the url argument in the joinRoom function.`,
```
AHGIJMKLKKZNPJKQR, line 47: **suggestion:** 
We could link to fishjam.io
```suggestion
          `You haven't passed your Fishjam ID  to the FishjamProvider, nor have you passed the url argument in the joinRoom function.`, You can get your Fishjam ID at https://fishjam.io/app
```

packages/react-client/src/FishjamProvider.tsx
czerwiukk: This is something I already know we need to work out, but I haven't done much thinking in this regard. +1

packages/react-client/src/hooks/useConnection.ts
czerwiukk: yeah, feels beneficial

packages/react-client/src/FishjamProvider.tsx
czerwiukk: Fishjam ID and Fishjam Management Token is the output of the offline discussion

packages/react-client/src/FishjamProvider.tsx
czerwiukk: we will leave it as is for now, though the error messages are clear about how to get and set the fishjam id

packages/react-client/src/hooks/useSandbox.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const useSandbox = (props?: UseSandboxProps) => {
+  const fishjamId = useContext(FishjamIdContext);
+
+  if (!fishjamId && !props?.configOverride) {
+    throw Error(`You haven't passed the fishjamId to the FishjamProvider.`);
+  }
+
+  const overridenFishjamUrl = props?.configOverride?.fishjamUrl;
+  const fishjamUrl = `${FISHJAM_HTTP_CONNECT_URL}/${fishjamId}`;
```
dawidmatyjasik, line 27: We should avoid using raw strings for `roomType`. It would be better to define a proper type to ensure type safety and avoid typos

Additionally, the current default value `conference` doesn't match the documentation:

https://docs.fishjam.io/room-manager#how-do-i-use-it

czerwiukk: d

dawidmatyjasik (approved):
GJ!

PR #386, opened 2025-07-21
Add livestream streamer API

AHGIJMKLKKZNPJKQR:
## Description

Added `useLivestreamStreamer`, which allows the user to start streaming with WHIP.

The hook accepts two MediaStreams, as that seems to be the most flexible API, and it integrates with the other hooks like `useCamera`, `useScreenShare` etc.

## Types of changes

      not work as expected)


packages/react-client/src/hooks/useLivestreamStreamer.ts
```diff
@@ -0,0 +1,67 @@
...
+    /** The audio source to publish. e.g. `microphoneStream` from {@link useMicrophone} or `stream` from {@link useScreenShare} */
+    audio: MediaStream;
+  };
+  /** Streamer token used to authenticate with Fishjam */
+  token: string;
+};
+
+/** @category Livestream */
+export interface UseLivestreamStreamerResult {
+  /**
```
czerwiukk, line 20: **suggestion:** as mentioned, `isConnected` flag could be useful here

packages/react-client/src/hooks/useSandbox.ts
```diff
@@ -11,6 +11,12 @@ type RoomManagerResponse = {
   peer: BasicInfo;
 };
 
+class FishjamIdMisconfiguredError extends Error {
```
czerwiukk, line 14: **praise:** :like:

examples/react-client/livestreaming/src/components/LivestreamViewer.tsx
```diff
@@ -17,10 +18,20 @@ import { Input } from "./ui/input";
...
+  roomName: streamerRoomName,
+}) => {
   const { connect, disconnect, stream, error } = useLivestreamViewer();
   const { getSandboxViewerToken } = useSandbox();
-  const [roomName, setRoomName] = useState<string>("");
+  const [nameOverriden, setNameOverriden] = useState<boolean>(false);
+  const [roomName, setRoomName] = useState<string>(streamerRoomName);
+
+  if (!nameOverriden && roomName != streamerRoomName)
+    setRoomName(streamerRoomName);
```
czerwiukk, line 34: **issue:** Setting a state in rendering phase is a bad practice and can lead to excess and even infinite renders. Wrap it in an useEffect to ensure the side effects are run in safe manner.

czerwiukk (changes requested):

packages/react-client/src/hooks/useLivestreamStreamer.ts
```diff
@@ -0,0 +1,84 @@
...
+    setIsConnected(false);
+  }, []);
+
+  const connect = useCallback(
+    async ({ inputs: { video, audio }, token }: ConnectStreamerConfig, urlOverride?: string) => {
+      if (resultRef.current !== null) disconnect();
+
+      const videoTrack = video?.getVideoTracks().at(0);
+      const audioTrack = audio?.getAudioTracks().at(0);
+      const stream = new MediaStream([videoTrack, audioTrack].filter((v) => v != null));
```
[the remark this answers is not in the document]
AHGIJMKLKKZNPJKQR: ts issue

czerwiukk (approved):

PR #523, opened 2026-04-24
fix(react-client): clear VAD outline on peer reconnect

Critteros:
## Description

Rewrite `useVAD` to derive each peer's speaking boolean directly from the current microphone tracks instead of a per-`trackId` cache. `voiceActivityChanged` now only drives a re-render via a version counter.

This is a crude fix scoped to `useVAD`. The underlying invariant — that a track's last observed `vadStatus` is `silence` when it disappears — is still not enforced at the webrtc-client layer, so other consumers of `FishjamTrackContext.vadStatus` could hit the same class of bug.

## Motivation and Context

When a peer disconnected mid-speech, their track was removed without a final `silence` vadNotification, leaving a stale `oldTrackId -> "speech"` entry in `useVAD`'s cache. On reconnect the new track wrote under a new `trackId`, so the OR-reduce `Object.values(tracks).some(v => v === "speech")` kept returning `true` and the talking-indicator outline stayed stuck on for every other viewer. Deriving from the live track list auto-prunes vanished tracks, so there is no state left to go stale.

## Documentation impact

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/packages/react-client/src/hooks/useVAD.ts
+++ b/packages/react-client/src/hooks/useVAD.ts
@@ -1,8 +1,7 @@
-import type { FishjamTrackContext, VadStatus } from "@fishjam-cloud/ts-client";
-import { useContext, useEffect, useMemo, useState } from "react";
+import { useContext, useEffect, useMemo, useReducer } from "react";
 
 import { FishjamClientStateContext } from "../contexts/fishjamState";
-import type { PeerId, TrackId } from "../types/public";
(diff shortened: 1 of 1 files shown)


MiloszFilimowski (approved):

MiloszFilimowski:
Fixes: FCE-3199

PR #526, opened 2026-05-06
FCE-3288 use current callback references

czerwiukk:
## Description

Introduces `useCurrentCallback` hook that ensures the closure values are always latest.

## Motivation and Context

Invoking methods one after another in `useEffect` was prone to stale closure issue.

## Documentation impact

## Types of changes

      not work as expected)

### DIFF
--- a/packages/react-client/src/hooks/internal/useCurrentCallback.ts
+++ b/packages/react-client/src/hooks/internal/useCurrentCallback.ts
@@ -1,0 +1,15 @@
+import { useCallback, useLayoutEffect, useRef } from "react";
+
+/**
+ * Returns a stable function reference whose body always reads the latest
+ * closure values. Use when a method needs to be called from effects or
+ * captured by consumers without being memoized against changing state
+ * (peerStatus, deviceTrack, etc.) in its dep array.
+ */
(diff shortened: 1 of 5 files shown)


packages/react-client/src/hooks/internal/useCurrentCallback.ts
```diff
@@ -0,0 +1,19 @@
+import { useCallback, useEffect, useLayoutEffect, useRef } from "react";
+
+// Falls back to useEffect on the server so SSR doesn't warn about
+// useLayoutEffect having no effect there.
+const useIsomorphicLayoutEffect = typeof window === "undefined" ? useEffect : useLayoutEffect;
```
MiloszFilimowski: **question:** Does it make sense to even account for? We use `useCurrentCallback` for purely web-based API, so SSR won't work here anyway.

czerwiukk: used it not to raise warnings on ssr, after hydration it's gonna be layout effect

czerwiukk: but you know what - i think let's not worry in advance and fix one at a time

MiloszFilimowski (approved):

PR #536, opened 2026-05-13
make useSandbox tolerate undefined props

czerwiukk:
`useSandbox()` now raises a TypeError, we can avoid that and raise the MissingSandboxApiUrlError later


Magmusacy (approved):
potat

PR #543, opened 2026-06-08
add unit hook tests

czerwiukk:
## Description

Adds unit tests for the react client level public hooks.

## Motivation and Context

Introducing hook tests that capture the current state of features will make moving logic to the typescript client easier. The goal is to move the features to the typescript client level without changing the behavior of the hooks.

## Documentation impact

## Types of changes

      not work as expected)


packages/react-client/src/tests/support/fakeFishjamClient.ts
```diff
@@ -0,0 +1,322 @@
...
+    this.emit("connectionStarted");
+    this.status = "initialized";
+    return new Promise<void>((resolve, reject) => {
+      const onSuccess = () => {
+        cleanup();
+        resolve();
+      };
+      const onError = () => {
+        cleanup();
+        reject();
```
MiloszFilimowski: `reject()` with no argument: any future test that hits `joinError`/`authError` during a pending `joinRoom` will fail with "unhandled rejection: undefined" — no signal about which event fired. Cheap insurance to reject with an `Error` naming the event, e.g. pass the event name into `onError`.

packages/react-client/src/tests/support/fakeMediaStream.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  addTrack(track: MediaStreamTrack): void {
+    if (!this.tracks.includes(track)) this.tracks.push(track);
+  }
+
+  removeTrack(track: MediaStreamTrack): void {
+    this.tracks = this.tracks.filter((t) => t !== track);
+  }
+
+  clone(): MediaStream {
```
MiloszFilimowski, line 51: Real `MediaStream.clone()` clones the tracks (new ids); this shares track instances. Nothing in the SDK calls `.clone()` today, but since this kit is meant to outlive the rewrite, a silently-wrong clone is worse than either cloning the tracks or `throw new Error("not implemented")`.

packages/react-client/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    await act(async () => {
+      // connect() resolves only once joined fires (matching the real client).
+      const joining = result.current.joinRoom({ peerToken: "tok-123", peerMetadata: { name: "alice" } });
+      client.simulateJoined();
+      await joining;
+    });
+
+    expect(client.connect).toHaveBeenCalledTimes(1);
+    const config = client.connect.mock.calls[0][0] as { url: string; token: string; peerMetadata: unknown };
```
MiloszFilimowski: This cast (and the one in the `peerMetadata` test below) exists because the fake types `connect`'s parameter as `unknown`. Typing it as the real `ConnectConfig` would drop the casts and extend the drift-tripwire idea to the config shape — the fake already imports types from `@fishjam-cloud/ts-client`.

MiloszFilimowski (approved):

packages/react-client/src/tests/screenShare.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    // The middleware output is pushed to the SFU via replaceTrack.
+    expect(client.replaceTrack).toHaveBeenCalled();
+
+    // KNOWN QUIRK (FCE-3574): setTracksMiddleware never writes the middleware
+    // back into state, so `currentTracksMiddleware` stays null and the middleware
+    // is NOT re-applied on a subsequent startStreaming. Captured here so the
+    // rewrite has to make a deliberate decision to fix it (the assertion will
+    // flip when it does). See REWRITE_PLAN.md §5.
```
Tay0108: Where can I find this `REWRITE_PLAN.md`?

Tay0108 (approved):

packages/react-client/src/tests/screenShare.spec.ts
MiloszFilimowski: Probably some leftover from the initial plan. I'll remove it

PR #544, opened 2026-06-10
Send all batchable commands in sdp offer

roznawsk:
## Description

This allows for negotiating both audio and video at the same time.

## Motivation and Context

## Documentation impact

## Types of changes

      not work as expected)

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,4 +1,10 @@
+approvedGitRepositories: []
+
+enableScripts: true
+
 nodeLinker: node-modules
+
+npmMinimalAgeGate: 0
 
 npmScopes:
   fishjam-cloud:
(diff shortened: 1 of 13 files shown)


.yarnrc.yml
```diff
@@ -1,5 +1,11 @@
+approvedGitRepositories: []
+
+enableScripts: true
+
 nodeLinker: node-modules
 
+npmMinimalAgeGate: 0
```
czerwiukk, line 7: **issue:** what's the reason for age gate 0?

czerwiukk (approved):

.yarnrc.yml
roznawsk: It's what generates by default when running yarn install with new version

PR #561, opened 2026-07-09
Add Android integration to voip calls

Magmusacy:
## Description

Added android side of voip-calls

## Motivation and Context

Now both users from android and ios can access voip-calls functionality

## Documentation impact

- [X] Documentation update required
- [ ] Documentation updated [in another PR](_)
- [ ] No documentation update required

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      not work as expected)

PR #562, opened 2026-07-16
[FCE-3435] Add call signaling handshake, native timeouts, hold, call waiting and Recents redial to VoIP calls

Magmusacy:
## Description

Disclaimer: for now I'm using expo plugins to configure appDelegate on iOS but I will fix it to use our wrapper package in future PR

Builds on the Android integration from #561 with the remaining call-lifecycle and native-call features:

- **WebSocket call signaling** (`useCallSignaling`, server support) with an answer-to-connect handshake on iOS and Android — the native call UI only reports the call as connected once the callee has actually joined, with a `connecting` state in between (`fulfillIncomingCallConnected` / `failIncomingCallConnected`, `reportOutgoingCallConnected`).
- **Native call timeouts** for incoming, outgoing, and answer-fulfillment phases, configurable via new plugin options on iOS and Android.
- **Hold / resume** (`setCallHeld`, `isCallHeld`) and **call waiting** for a second incoming call (via a `react-native-webrtc` submodule bump).
- **Native mute surfaced to JS** and end-call events with reasons, so the app stays in sync with CallKit/Telecom actions.
- **`enableCallIntents` plugin option** so calls appear in the system Recents and can be redialed from there (iOS).
- **Example polish** — Fishjam branding (icons, splash), server-served fish avatars, `DirectoryScreen` → `UsersScreen`, speaker toggle fixes, and camera PIP improvements.

## Motivation and Context

#561 made basic calls work on both platforms, but the example didn't yet behave like a real phone app: answering connected immediately even if signaling later failed, unanswered calls rang forever, there was no hold or second-call handling, and muting from the native call UI desynced the app. This PR closes those gaps so the example serves as a complete reference for production VoIP behavior on Fishjam.

## Documentation impact

## Types of changes

      not work as expected)


examples/mobile-client/voip-call/app/src/signaling/useCallSignaling.ts
```diff
@@ -0,0 +1,112 @@
...
+      // Caller cancelled an outgoing call that was still ringing.
+      if (prevStatus === 'connecting' && prevCall.isOutgoing) {
+        sendSignal({
+          type: 'call-cancelled',
+          to: prevCall.handle,
+          roomName: prevCall.roomName,
+        });
+      }
+      // Callee rejected an incoming call before answering.
+      else if (prevStatus === 'incoming' && !prevCall.isOutgoing) {
```
MiloszFilimowski, line 109: **issue:** You send `call-rejected` for every incoming call that ends before answer, but this effect can't tell why it ended. When the native incoming timeout fires (`missed`), or when we end because we received `call-cancelled`, you still send `call-rejected` — the caller's call then ends with `remote`, so it looks like I declined when I just didn't pick up. We should check `lastEndedReason` here and only signal for the user-initiated end.

examples/mobile-client/voip-call/app/src/voip/VoipProvider.tsx
```diff
@@ -97,116 +161,385 @@ export function VoipProvider({
...
+            connectedRoom !== payload.roomName &&
+            currentCallRef.current != null;
+
+          if (isAcceptedWaitingCall) {
+            swapInProgressRef.current = true;
+          }
+
+          try {
+            if (isAcceptedWaitingCall) {
+              // The old call's `onEnded` may not have been processed yet, so tear
```
MiloszFilimowski: **question:** If this path is real and old call's `onEnded` arrives after this `onIncoming`, the queued `onEnded` will run `endCall` with `currentCallRef` already pointing at the new call — `onEnded` has no room id, so it ends the call we just accepted. I checked the native side and `ended(old)` always comes before `incoming(new)` on both platforms, so maybe this guard is dead code? If it can happen, the late `onEnded` needs a way to know it belongs to the old call.

examples/mobile-client/voip-call/app/src/signaling/useCallSignaling.ts
Magmusacy: true, i didn't think it's that important in that example because the ability to distinguish other reject types is still here

examples/mobile-client/voip-call/app/src/voip/VoipProvider.tsx
Magmusacy: yes it's dead code

MiloszFilimowski (approved):

PR #566, opened 2026-07-20
[FCE-3464] Add ios-expo-voip package for iOS appDelegate config

Magmusacy:
## Description

Replaces the `withAppDelegate` regex block in the iOS config plugin with a proper Expo module.

- **New package `@fishjam-cloud/expo-voip`** — glue-only, iOS-only, no JS. One `ExpoAppDelegateSubscriber` that registers PushKit at launch and forwards `INStart*` call intents. Both gated on the `FishjamVoipEnabled` Info.plist flag.
- **`withFishjamIos.ts`** — regex block deleted. The plugin now just writes `FishjamVoipEnabled` and throws if `voip` options are set without the new package installed.
- **`react-native-webrtc` submodule bump** — `DEFINES_MODULE => YES` so the Swift subscriber can `import FishjamReactNativeWebrtc`.

No Info.plist keys are lost. Verified on device: the subscriber runs at launch, a VoIP token is issued, and the device registers.

## Motivation and Context

The old approach string-spliced generated Swift in `AppDelegate.swift`. It silently no-ops whenever Expo changes its template or the regex stops matching — and the failure mode is an app that builds and launches fine but can never receive a call. It also forced consumers to hand-add a bridging-header import and an early `registerForVoIPPushes()` call.

`ExpoAppDelegateSubscriber` is the supported extension point for this. It's re-applied by autolinking instead of patching generated code, so it survives prebuild, and both manual steps go away.

## Documentation impact

The VoIP README still documents the old flow (hand-editing `AppDelegate.swift`, bridging header) and wrongly claims `aps-environment` is already present and should be set in Xcode.

## Types of changes

      not work as expected)


packages/ios-expo-voip/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@fishjam-cloud/ios-expo-voip",
```
MiloszFilimowski, line 2: **issue:** There is no publish job for this package in `release.yaml`. The plugin now throws when `voip` options are set and this package is not installed, so after the next release users can't enable VoIP at all — the package won't exist on npm. We should add a publish job (with the same `check:webrtc-published` guard the other native packages use, since the pod depends on `FishjamReactNativeWebrtc`).

packages/mobile-client/plugin/src/withFishjamIos.ts
```diff
@@ -335,13 +329,33 @@ const withFishjamVoipTimeouts: ConfigPlugin<FishjamPluginOptions> = (config, pro
     return configuration;
   });
 
+const withFishjamExpoVoip: ConfigPlugin<FishjamPluginOptions> = (config, props) => {
+  if (!props?.voip) {
+    return config;
+  }
+
+  try {
+    require.resolve('@fishjam-cloud/ios-expo-voip/package.json');
```
MiloszFilimowski: **issue:** `require.resolve` here resolves from inside `react-native-client`, not from the app. In a monorepo where the SDK is hoisted to the root but the app installs `ios-expo-voip` in its own `node_modules`, this check fails even though the package is installed. We should resolve with `{ paths: [projectRoot] }`, `modRequest.projectRoot` is available inside the mod callback.

MiloszFilimowski (approved):

PR #571, opened 2026-07-23
[FCE-3647] Move room joining out of VoipProvider and into the app

Magmusacy:
## Description

`VoipProvider` is now only the native CallKit / Core-Telecom call state machine.
Joining and leaving the room, peer tokens and camera/mic lifecycle move to the app.

- Props reduced to `isVideo` and `onWaitingCallDeclined`. Removed `getPeerToken`,
  `requestCall`, `canStartOutgoingCall`.
- `useVoip()` gains `reportConnected()` / `reportConnectFailed()`, which the app calls
  when its own join succeeds or fails. This fulfills CallKit's answer action.
- `useVoip()` also gains `isMuted` and `pendingCallIntent` / `clearCallIntent()`.
- `answerCall` removed — an answer can only originate from the system call UI. (we only supported it anyway because we always shown the ssystem notification)
- `startCall(to, roomName)` only reports the call to CallKit; it no longer signals or joins.
- Hold and mute are reported as state instead of applied to tracks the SDK doesn't own.

Example app gains `src/voip/useVoipRoomConnection.ts` (join/leave, report connected,
apply hold and mute) and `src/voip/usePlaceCall.ts` (room name, signaling, `startCall`).

## Motivation and Context

Adopting VoIP previously meant handing your connection to the SDK: `VoipProvider` had to
live inside `FishjamProvider`, took over join/leave, and only offered a `getPeerToken`
hole to fill. Now a VoIP call connects through the same code as any other Fishjam session.

Fulfilling CallKit's answer action can't move, since the `requestId` is a native handle
only the SDK sees — hence the `reportConnected()` acknowledgement.

## Documentation impact

## Types of changes

      not work as expected)


examples/mobile-client/voip-call/app/src/screens/InCallView.tsx
```diff
@@ -42,10 +42,73 @@ function useElapsed(startedAt: number | null): number {
...
+    (async () => {
+      if (isOnHold) {
+        heldMediaRef.current = {
+          microphoneEnabled: isMicrophoneOn,
+          cameraEnabled: isCameraOn,
+        };
+        if (isMicrophoneOn) await toggleMicrophone();
+        if (isCameraOn) await toggleCamera();
+      } else {
+        const { microphoneEnabled, cameraEnabled } = heldMediaRef.current;
```
MiloszFilimowski, line 73: **issue:** Resume from hold restores the mic from `heldMediaRef` without looking at `isMuted`. Put the call on hold with mic on, then mute from the CallKit screen while it's held, then resume — mic goes back on but CallKit still shows muted. You are talking and the system UI says you are not. On iOS the hold happens by itself when a cellular call comes in, so this is not only a manual path.

MiloszFilimowski (approved):

PR #585, opened 2026-08-04
FCE-3578: add StateStore with synchronous notifications

MiloszFilimowski:
## Description

- Adds `StateStore` to `tsunami`: `getState` / `subscribe` / `subscribeToSlice`, one notification per effective `update()`, structural sharing (unchanged slices keep their references).
- Notifications are **synchronous** — this supersedes the microtask-batched version from the closed #581.
- Guards the tsunami tsconfig against source-adjacent emit (`noEmit` in base, off in build).

## Motivation and Context

The store from RFC 0015 §2–3 that the framework adapters subscribe to. Synchronous notify is required by react-client's connection specs, which assert inside synchronous `act()` — microtask batching broke them. Extracted from the validated reference on `spike/tsunami-full`.

## Documentation impact

- [ ] Documentation update required
- [ ] Documentation updated [in another PR](_)
- [x] No documentation update required

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      not work as expected)

PR #586, opened 2026-08-04
FCE-3578: wire connection state into the StateStore

MiloszFilimowski:
Stacked on #585.

## Description

- `ClientState` gets its connection slices: `peerStatus`, `reconnectionStatus`, `localPeer`, `remotePeers`, `components`.
- The `FishjamClient` wrapper mirrors session state into the `StateStore`: each signalling event composes everything it affects into **one** store update; participant slices are re-read with fresh references (ts-client mutates peers in place).
- `signallingClient` injection seam (`@internal`), adopted eagerly so injected fakes can emit before any method call — the seam the react-client suite drives.
- Public surface: `getState` / `subscribe` / `subscribeToSlice`; `dispose()` clears the store.

## Motivation and Context

Framework adapters (react-client next in the stack) need a subscribable, synchronously readable snapshot instead of hand-tracking signalling events. One-event-one-update keeps re-render counts sane.

## Documentation impact

## Types of changes

      not work as expected)

### DIFF
--- a/packages/tsunami/src/FishjamClient.ts
+++ b/packages/tsunami/src/FishjamClient.ts
@@ -99,7 +99,10 @@ export class FishjamClient<PeerMetadata = GenericMetadata, ServerMetadata = Gene
 
     // An injected signalling client can emit events before any client method
     // is called, so its event forwarding must be wired immediately.
-    if (signallingClient) this.adoptTsClient(signallingClient);
+    if (signallingClient) {
+      this.tsClient = signallingClient;
+      this.wireTsClient(signallingClient);
+    }
   }
 
   /** Synchronously readable snapshot of the client's observable state. */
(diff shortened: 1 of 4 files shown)


packages/tsunami/src/FishjamClient.ts
```diff
@@ -41,12 +85,37 @@ export class FishjamClient<PeerMetadata = GenericMetadata, ServerMetadata = Gene
   private readonly resources = new ClientResourceScope();
 
   private tsClient: TsClient<PeerMetadata, ServerMetadata> | null = null;
+  private readonly injectedTsClient: TsClient<PeerMetadata, ServerMetadata> | null = null;
```
czerwiukk: **suggestion:** why is the tsClient duplicated? the injectedTsClient is set as the tsClient in the `getTsClient` method, so why not simplify this?

MiloszFilimowski: Good point I overlooked it. It's a left over from the previous lazy design.

czerwiukk (approved):

PR #613, opened 2026-09-18
Publish and pause devices started or stopped while connected

Magmusacy:
Stacked on #612.

## Description

**`startMicrophone()` / `startCamera()` and `stopMicrophone()` / `stopCamera()` now update the published track while connected**, the same way `toggleMicrophone()` / `toggleCamera()` do.

- **Start while connected:** publishes the device, or replaces the published track if there is one.
- **Stop while connected:** stops the device immediately, then pauses the published track (`paused: true`).
- `toggleDevice` now reuses the same start and stop logic, so the two paths can't drift apart.
- Not connected: unchanged. A device started before joining is still published on `joined`; stopping after `leaveRoom()` sends nothing.

## Motivation and Context

Start and stop only touched the local device, so the room fell out of sync:

- **stop, then start:** the stopped track stayed published, so others heard silence for the rest of the call while the app showed the mic as on;
- **stop:** others still saw the device as not paused;
- **start after joining:** nothing was published.

On mobile, where muting is done by stopping the microphone, this made unmuting leave the user silent.

Checked:
- `startStopDeviceWhileConnected.spec.ts`: 7 tests fail before the change, all 10 pass after; the full react-client suite passes.
- On an Android emulator in a sandbox room: before, the sender kept the ended track after stop then start; after, stop paused the track and start put the live microphone back on the same published track.

## Documentation impact

## Types of changes

      not work as expected)


Magmusacy:
Parked for now, tracked in FCE-3857. Branch kept.

